PR #1, opened 2014-11-11
Correctly handle exclude patterns in git_hook

mpenkov:
The git hook wasn't handling exclude patterns correctly. This commit fixes that problem.


mpenkov:
Sorry it's taken me so long to reply. I've added some unit tests. I'll make another pull request.

PR #2, opened 2015-02-07
Correctly handle exclude patterns in git_hook (now with unit tests)

mpenkov:
The git hook wasn't handling exclude patterns correctly. These two commits fix that problem and add unit tests to demonstrate it.

### DIFF
--- a/flake8/hooks.py
+++ b/flake8/hooks.py
@@ -16,6 +16,11 @@ from flake8.engine import get_parser, get_style_guide
 from flake8.main import DEFAULT_CONFIG
 
 
+def fix_exclude(flake8_style, tmpdir):
+    flake8_style.options.exclude = [tmpdir + x if "/" in x else x
+                                    for x in flake8_style.options.exclude]
+
+
 def git_hook(complexity=-1, strict=False, ignore=None, lazy=False):
     """This is the function used by the git hook.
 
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Hey @mpenkov !

Thanks for the tests! It seems like they aren't working exactly like you intended. If you look at the pull request you can see that [Travis CI ran the tests and they failed][testrun]. I wonder if you can write a test (without the `fix_exclude` function) that shows that it behaves correctly now. If you could just explain the old behaviour as a comment here, that would be helpful.

Also, for what it is worth, you don't need to keep opening new pull requests. If you look at #1 you'll see that it has your new commit d39fd7d there too. So just keep pushing to master and the new code will just appear here without having to close old PRs and open new ones.

Thanks,
Ian

[testrun]: https://travis-ci.org/PyCQA/flake8/builds/49836075

mpenkov:
Hi Ian,

Yes, you're right. It's behaving completely the opposite of what I intended - it's actually breaking the hook. I'm not sure why this happens - perhaps a difference in the environment (I'm on a Mac) is contributing to the cause. Both tests are passing on my side.

Is there a way I can run this in the same environment as Travis CI without pushing commits?

The old behavior, as I'm observing on my machine, is as follows. Assume we have a project with two files: myproject/main.py and myproject/test/test.py. Given an exclude pattern myproject/test/test.py, the flake8 style returned by get_style_guide has the exclude pattern /Users/misha/git/flake8/myproject/test/test.py (the absolute path to the actual existing test.py file). When we run the git hook, it will copy myproject/main.py and myproject/test/test.py to some temporary directory (e.g. /tmp/Users/misha/git/flake8/myproject/main.py and /tmp/Users/misha/git/flake8/myproject/test/test.py, respectively), and then attempt to match those paths to the exclude pattern. For test.py, the match will fail, since the path /tmp/Users/misha/... does not match the pattern /Users/misha/...). Since the match fails, test.py won't be excluded, contrary to our intentions.

sigmavirus24:
@mpenkov :sparkles: :cake: :sparkles: Thank you!

Please feel free to send another Pull Request to add yourself to CONTRIBUTORS.txt

PR #8, opened 2016-11-09
Implemented output to JUnit XML file

dirkmoors:


sigmavirus24:
Hi @dirkmoors!

Thanks for taking the time to send your pull request! A few things to start:

1. All pull requests must pass the tests (without deleting the tests) before they can be merged
2. Flake8 is specifically designed to allow people to write custom plugins to report the output as they wish. These are documented [here](http://flake8.pycqa.org/en/latest/plugin-development/formatters.html). You can also read the source of our [default](https://gitlab.com/pycqa/flake8/blob/master/src/flake8/formatting/default.py) formatting plugins here.
3. Finally, this project prefers to discuss all new features prior to implementing them via a pull/merge request. 

You should consider turning this into a plugin that you distribute independently from Flake8. Once that plugin has matured, we can [discuss](https://gitlab.com/pycqa/flake8/issues) including it by default.

Cheers!
Ian

dirkmoors:
@sigmavirus24 

Thank you for your feedback!

I'm familiar with the plugin architecture, but at the moment, it lacks the possibility to format 'successful' files, only files that contain issues.

When you're working in a CI environment (e.g. Bamboo), you need a JUnit XML format output to integrate better into the CI. The idea is that you output all processed files, (even the ones that did not fail). That allows the CI to show all processed files.

How would you advise me to achieve this functionality instead?

Thank you in advance!
Dirk

dirkmoors:
This is what I'm trying to achieve:

image

sigmavirus24:
> How would you advise me to achieve this functionality instead?

It would require a bit of feature work. It's not out of the question, but this discussion belongs on GitLab.

dirkmoors:
@sigmavirus24 Ok thanks! I've created a ticket here: https://gitlab.com/pycqa/flake8/issues/251

PR #9, opened 2016-11-22
Require mock 2.0.0, because unit tests are failing on older one

onovy:


sigmavirus24:
Where are you getting an older version of mock?

onovy:
for example in Debian Jessie is mock 1.0.1

sigmavirus24:
Please update https://github.com/PyCQA/flake8/blob/a981fe34ed5d02c762e40e1ed2d6f309290035b3/tox.ini#L7 too. I understand downstream packagers don't care about tox.ini but there's no need for the two to be out of sync.

onovy:
right, fixed

sigmavirus24:
Thanks @onovy!

onovy:
Thanks too

PR #10, opened 2017-02-02
Use current Python on flake8, pylint, docs, bandit

cclauss:


sigmavirus24:
PyLint doesn't support Python 3.6 yet. Further, you've not explained *why* we should change those versions. We do not change things simply for the sake of changing them.

PR #14, opened 2017-11-24
fix for flake8-isort

r3m0t:
To test this, create a setup.cfg with:

```
[flake8]
accept-encodings = utf-8
max-line-length = 119

[isort]
line_length=119
```

install `flake8-isort`, install the git hook, and write a python file with a long import line.

`isort` run separately doesn't wrap the line, but the git hook complains because isort can't find `setup.cfg` in the temporary directory.


r3m0t:
https://gitlab.com/pycqa/flake8/merge_requests/215

PR #15, opened 2017-12-11
Use dash-seperated name for PYPI package

5j9:
The name of the package will be normalized by `setuptools` anyway
and it will most likely appear as dash-separated on PYPI.[1]

This change also helps to keep the distinction between the PYPI
package name  and regular python packages (the directories with
__init__.py files).

[1]: https://mail.python.org/pipermail/distutils-sig/2011-August/017936.html

### DIFF
--- a/docs/source/plugin-development/registering-plugins.rst
+++ b/docs/source/plugin-development/registering-plugins.rst
@@ -34,13 +34,13 @@ like:
     flake8_entry_point = # ...
 
     setuptools.setup(
-        name="flake8_example",
+        name="flake8-example",
         license="MIT",
         version="0.1.0",
         description="our extension to flake8",
         author="Me",
         author_email="example@example.com",
-        url="https://gitlab.com/me/flake8_example",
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Hi @5j9. This is just some example code, can you explain what about this change feels necessary to you?

5j9:
Hi, @sigmavirus24!
Well, I followed this example to create a plugin and used `_`. Everything worked fine, except that after I uploaded it to PYPI the package name became dash-separated. I didn't like the discrepancy between that and my setup.py and repo name. So I renamed my repo and amended the setup.py file.
I thought that fixing this example could have prevented me from doing this extra step... However, maybe I did a stupid thing that no one else does, or maybe it's not that important and I should have just ignored it. So it's OK if the change will not get accepted, think of it as just a suggestion. Thanks for the great module anyway!

danpalmer:
The test failures here look unrelated, I reckon this is good to merge?

PR #18, opened 2018-02-20
enable pip cache in appveyor build

thijstriemstra:
The reason for this pull request is the same reason caching was added to pip: we want to minimze the amount of traffic and load on the pypi site. Downloading any package(s) from pypi more than once is a waste of resources.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,5 +4,8 @@ install:
 
 build: off
 
+cache:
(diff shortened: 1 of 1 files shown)


thijstriemstra:
Can anyone let me know this trivial change can be merged?

thijstriemstra:
Ok, nevermind.

PR #23, opened 2018-08-20
Add --prepend-config command line option

gchazot:
Add code to handle config files specified with the --prepend-config
argument.

Options specified in prepend files get a higher precedence than flake8
defaults but a lower one than user configs. This allows to override
the defaults without affecting the behaviour of user and local config
files.

This implements a solution for issue #349


gchazot:
Moved this to GitLabs as it is the true home of the project: https://gitlab.com/pycqa/flake8/merge_requests/252

PR #25, opened 2018-10-08
Preserve checkers which will not be processed

robert-b-clarke:
Ensure that errors captured during the creation of checkers, for instance a non existent file or directory, are preserved and reported.

Currently it is possible to invoke flake8 with a non existent filepath, and execution will proceed without raising any errors, or a non zero exit code. Although the error is caught in `checker.FileChecker._make_processor` it is unpreserved

This fix addresses the issue by persisting checkers that never get run and including them in the final report

### DIFF
--- a/src/flake8/checker.py
+++ b/src/flake8/checker.py
@@ -76,6 +76,7 @@ class Manager(object):
         self.using_multiprocessing = self.jobs > 1
         self.processes = []
         self.checkers = []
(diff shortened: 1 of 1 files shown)


robert-b-clarke:
apologies, I've just seen that the build failed. I will fix and resubmit

robert-b-clarke:
Fixed now (overwrites previous commit)

PR #29, opened 2018-11-02
FIX: parsing of per-file-ignores

tacaswell:
Tested on Matptoltilbs config https://github.com/matplotlib/matplotlib/blob/bcb372b57367b3c6ad9973eae14a72800b97324f/.flake8#L24-L261

Can you point me to what sort of tests you want on this and where to
add it?


tacaswell:
Never mind, this is very incorrect

tacaswell:
OK, pushed an entirely different commit that actually works .

PR #30, opened 2018-11-03
DOCS: Add clarification to VCS checks

jmcgill298:


jmcgill298:
This is a simple update to docs, but I spent an hour or so debugging why it was finishing the commit when I know there are files in the project that are not compliant. If you have other doc updates, I don't mind closing this PR and letting this update or a similar update be part of that change.

PR #32, opened 2019-01-18
Add support for hooking flake8 into the post-commit hook of Breezy (successor of Bazaar).

jelmer:
Add support for hooking flake8 into the post-commit hook of Breezy (successor of Bazaar).

This makes flake8 ship a plugin for Breezy. "flake8 --install-hook=breezy" sets
a flag in the repository that enables the plugin.

The only other variable is 'flake8.strict' (set in the local or per-repository
breezy config), which determines (like with Git) whether flake8 failures
cause the commit to fail.


jelmer:
Instead, we have now added this support to Breezy.

PR #36, opened 2019-03-08
Fixes handling of empty lists by Application

charlesfrye:
`Application.parse_preliminary_options_and_args` was previously, against
expectations, treating empty lists passed as the `argv` argument the
same way it treated `None`s.

This has been addressed and the correct behavior tested for in a unit
test of the `Application` class.

See issue #518 for details.

### DIFF
--- a/tests/unit/test_application.py
+++ b/tests/unit/test_application.py
@@ -1,5 +1,6 @@
 """Tests for the Application class."""
 import optparse
+import sys
 
 import mock
 import pytest
@@ -97,3 +98,12 @@ def test_prelim_opts_args(application):
     assert application.prelim_opts.statistics
     assert application.prelim_opts.verbose
     assert application.prelim_args == ['src', 'setup.py']
+
+
+def test_prelim_opts_handles_empty(application):
+    """Verify empty argv lists are handled correctly."""
+    irrelevant_args = ['myexe', '/path/to/foo']
+    with mock.patch.object(sys, 'argv', irrelevant_args):
+        application.parse_preliminary_options_and_args([])
+
+        assert application.prelim_args != irrelevant_args
--- a/tests/unit/test_application.py
+++ b/tests/unit/test_application.py
@@ -106,4 +106,4 @@ def test_prelim_opts_handles_empty(application):
     with mock.patch.object(sys, 'argv', irrelevant_args):
         application.parse_preliminary_options_and_args([])
 
-        assert application.prelim_args != irrelevant_args
+        assert application.prelim_args == []
--- a/src/flake8/main/application.py
+++ b/src/flake8/main/application.py
@@ -122,7 +122,7 @@ class Application(object):
         # do not need to worry and we can continue. If it is, we successfully
         # defer printing the version until just a little bit later.
         # Similarly we have to defer printing the help text until later.
-        args = (argv or sys.argv)[:]
+        args = (argv if argv is not None else sys.argv)[:]
         try:
             args.remove("--version")
         except ValueError:


charlesfrye:
@asottile here it is! I was able to check that my test and the rest of `test_application.py` passed, but I ran into problems testing the rest.

asottile:
Awesome, I'll take a look at it in the morning with a clearer head :)

tests/unit/test_application.py
```diff
@@ -97,3 +98,12 @@ def test_prelim_opts_args(application):
...
     assert application.prelim_args == ['src', 'setup.py']
+
+
+def test_prelim_opts_handles_empty(application):
+    """Verify empty argv lists are handled correctly."""
+    irrelevant_args = ['myexe', '/path/to/foo']
+    with mock.patch.object(sys, 'argv', irrelevant_args):
+        application.parse_preliminary_options_and_args([])
+
+        assert application.prelim_args != irrelevant_args
```
asottile: let's assert here that they're `== []` instead of `!= ...` (that way we're sure we have the right values)

Let's also add a test that makes sure:

```diff
+        application.parse_preliminary_options_and_args(['f1', 'f2'])
+
+        assert application.prelim_args == ['f1', 'f2']
```

I think we'll have to adjust to [`sys.argv[1:]`](https://github.com/PyCQA/flake8/blob/16ca17388ab4a8136617390329328e1a13219f84/src/flake8/main/application.py#L125) and [remove the `[1:]`](https://github.com/PyCQA/flake8/blob/16ca17388ab4a8136617390329328e1a13219f84/src/flake8/main/application.py#L141) after parsing -- might need to clean up a few other callsites as well

sigmavirus24: What other callsites are you thinking of? I don't see us slicing out the executable name anywhere but in the preliminary handling

asottile: I think only the three in these calls to `.initialize(` in these tests:
- https://github.com/PyCQA/flake8/blob/88caf5ac484f5c09aedc02167c59c66ff0af0068/tests/integration/test_plugins.py#L47

charlesfrye: > let's assert here that they're == [] instead of != ... (that way we're sure we have the right values)

Can do. Regarding the positive test, I think that's handled above, in `test_prelim_opts_args` on line 100.

Happy to also change the location of the slicing, but I don't think I understand what needs to change with the other callsites. In the calls to `.initialize`, they seem to include the executable name, which is what `parse_preliminary_options_and_args` expects. No one is directly talking to `parse_known_args` under the options manager.

asottile: you're right, that case is tested above

ah this conversation is confusing because half of it happened on github.

There's other callsites which call `initialize(...)` and don't pass the executable (which are currently also broken):

- https://gitlab.com/pycqa/flake8/blob/88caf5ac484f5c09aedc02167c59c66ff0af0068/src/flake8/main/git.py#L46
- https://gitlab.com/pycqa/flake8/blob/88caf5ac484f5c09aedc02167c59c66ff0af0068/src/flake8/main/mercurial.py#L39

My suggestion would be to use this patch and then fix the callsites, though it might make more sense to just fix the `.initialize(...)` callers:

```diff
diff --git a/src/flake8/main/application.py b/src/flake8/main/application.py
index c7859b9..f936054 100644
--- a/src/flake8/main/application.py
+++ b/src/flake8/main/application.py
@@ -122,7 +122,7 @@ class Application(object):
         # do not need to worry and we can continue. If it is, we successfully
         # defer printing the version until just a little bit later.
         # Similarly we have to defer printing the help text until later.
-        args = (argv if argv is not None else sys.argv)[:]
+        args = (argv if argv is not None else sys.argv[1:])[:]
         try:
             args.remove("--version")
         except ValueError:
@@ -137,8 +137,7 @@ class Application(object):
             pass
 
         opts, args = self.option_manager.parse_known_args(args)
-        # parse_known_args includes program name and unknown options as args
-        args = [a for a in args[1:] if not a.startswith("-")]
+        args = [a for a in if not a.startswith("-")]
         self.prelim_opts, self.prelim_args = opts, args
 
     def exit(self):
```

charlesfrye: In e9b9ebb, (as of writing, undergoing CI checks) I did not include the changes to slicing. Was that part of what you wanted to include in this PR?

asottile:
Moved the PR to gitlab: https://gitlab.com/pycqa/flake8/merge_requests/310

tests/unit/test_application.py
asottile: I can handle that in a followup, I realized after reading through the code a bit more that that's orthogonal

PR #38, opened 2019-09-12
Print errors to stderr

blueyed:
E.g.

    There was a critical error during execution of Flake8:
    Expected `per-file-ignores` to be a mapping from file exclude patterns to ignore codes.

    Configured `per-file-ignores` setting:

        …


joaoe:
Question: did you look for more uses of `print()` ?
PS: I'm not affiliated with the project.

blueyed:
@joaoe no, but would be easy to do when this gets accepted in general.

Issue #740, opened 2021-04-03
Wrong location reported for syntax errors


asottile:
In GitLab by @scascketta on May 21, 2020, 14:14

*Please describe how you installed Flake8*

```
$ python3.8 -m venv .venv
$ source .venv/bin/activate
(.venv) $ pip install flake8
```

*Please provide the exact, unmodified output of `flake8 --bug-report`*
```
{
  "dependencies": [],
  "platform": {
    "python_implementation": "CPython",
    "python_version": "3.8.1",
    "system": "Darwin"
  },
  "plugins": [
    {
      "is_local": false,
      "plugin": "mccabe",
      "version": "0.6.1"
    },
    {
      "is_local": false,
      "plugin": "pycodestyle",
      "version": "2.6.0"
    },
    {
      "is_local": false,
      "plugin": "pyflakes",
      "version": "2.2.0"
    }
  ],
  "version": "3.8.1"
}
```
*Please describe the problem or feature*

Running flake8 on a Python module with a syntax error produces the E902 TokenError with the wrong line and column reported.

*If this is a bug report, please explain with examples (and example code) what you expected to happen and what actually happened.*

The contents of the test file with invalid syntax:
```
(.venv) $ cat blah.py
print("hello, world")
)
```

Python reports the correct location of the syntax error:
```
(.venv) $ python3.8 blah.py
  File "blah.py", line 2
    )
    ^
SyntaxError: unmatched ')'
```

Running flake8 on it reports the wrong line and column of the syntax error:
```
(.venv) $ flake8 blah.py
blah.py:1:1: E902 TokenError: EOF in multi-line statement
```

Thank you for reading!

asottile:
In GitLab by @asottile on May 21, 2020, 14:23

tokenization is done before ast parsing so the errors are sometimes different and on different positions:

```console
$ python3 -m tokenize
print("hello, world")
)
1,0-1,5:            NAME           'print'        
1,5-1,6:            OP             '('            
1,6-1,20:           STRING         '"hello, world"'
1,20-1,21:          OP             ')'            
1,21-1,22:          NEWLINE        '\n'           
2,0-2,1:            OP             ')'            
2,1-2,2:            NEWLINE        '\n'           
<stdin>:3:0: error: EOF in multi-line statement
```

though it looks like `TokenError` is always given `1:1` and could maybe be slightly improved (though `3:0` in that example is *also* out of bounds)

asottile:
[...]

PR #1311, opened 2021-04-08
Fix wrong location reported for syntax errors

vanschelven:
See #740

This solves the immediate problem of hiding relevant line numbers, by showing both E902 and E999. Admittedly it would be even better if only the E999 would show up in that case... I went for the 90% solution here though.

I did not provide new tests, but rather adapted the 2 existing tests to reflect the new (and desirable) behavior.

### DIFF
--- a/src/flake8/checker.py
+++ b/src/flake8/checker.py
@@ -580,8 +580,8 @@ class FileChecker:
         """Run checks against the file."""
         assert self.processor is not None
         try:
-            self.process_tokens()
             self.run_ast_checks()
+            self.process_tokens()
         except exceptions.InvalidSyntax as exc:
             self.report(
                 exc.error_code,
--- a/tests/integration/test_main.py
+++ b/tests/integration/test_main.py
@@ -1,6 +1,8 @@
 """Integration tests for the main entrypoint of flake8."""
 import json
 import os
+import platform
+import sys
 from unittest import mock
 
 import pytest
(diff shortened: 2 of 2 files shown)


vanschelven:
I'm having some trouble understanding what's going on here (not a flake8 expert by any means) but here's what I found:

I appears that on of the tests that I wrote (on a Python 3.8 system) are not working on older versions of Python (3.6 and 3.7). In particular,

`test_tokenization_error_but_not_syntax_error` is missing the `"1:8: E999 SyntaxError: unexpected EOF while parsing"` that I just added.

The fact that it is missing this, however, would appear to be the expected behavior if we look at the name of the test ("but not syntax error")... which leads me to believe that the fact that this particular scenario is not a syntax error is no longer true since Python 3.8, and that the way forward is to repair this test in another way.

I'm a bit out of my depth other than the above though.

asottile:
the syntax error is different across versions because I fixed the underlying bug in the C parser to make it in line with the python parser

vanschelven:
@asottile what would you recommend going forward on this PR (if anything)?

asottile:
> @asottile what would you recommend going forward on this PR (if anything)?

the easiest fix is to assert different things based on different versions

vanschelven:
I think I'm giving up here... too much "stuff" for me to get right in a drive-by fix

asottile:
picked this up in #1320 if you want to try that branch out

Issue #1274, opened 2021-04-03
Add --extend-select command line argument - [opened]


asottile:
In GitLab by @fwinkl on Jun 8, 2020, 07:22

_Merges extend_select -> master_

Implement `--extend-select` command line argument following what was
done for `--extend-ignore` in !233. This option can be used to
selectively add individual codes without overriding the default list
entirely.

Addresses the remaining item of issue #365.

asottile:
In GitLab by @fwinkl on Jun 8, 2020, 07:24

Commented on [tests/unit/test_decision_engine.py line 198](https://github.com/PyCQA/flake8/compare/a7be77f761a4c29121d6bb6f61c11902281f9105..7db4f246a7b6c430baaac9fbef32698caa879d16#diff-f99de482e9d9043eff5fd016859a1862R198)



The changes here (`extend_select` -> `extended_default_select`) are just to avoid confusion with the now separate `extend_select` command line option).

asottile:
In GitLab by @fwinkl on Jun 8, 2020, 07:26

Not sure why the docs unit test fails. `tox -e docs` works locally for me and I did not make any changes in `FileProcessor`. Could someone advise?

asottile:
In GitLab by @asottile on Jun 10, 2020, 07:44

I've banned sphinx==3.1.0 -- they committed a regression

asottile:
In GitLab by @fwinkl on Jun 10, 2020, 23:56

added 4 commits

<ul><li>ea753d3c...a7be77f7 - 3 commits from branch <code>pycqa:master</code></li><li>7db4f246 - Add --extend-select command line argument</li></ul>

[Compare with previous version](/pycqa/flake8/-/merge_requests/437/diffs?diff_id=95128657&start_sha=ea753d3c5405ce5791d689f89168e3e83954efa2)

asottile:
In GitLab by @codecov on Jun 10, 2020, 23:57

# [Codecov](https://codecov.io/gl/pycqa/flake8/pull/437?src=pr&el=h1) Report
> Merging [#437](https://codecov.io/gl/pycqa/flake8/pull/437?src=pr&el=desc) into [master](https://codecov.io/gl/pycqa/flake8/commit/a7be77f761a4c29121d6bb6f61c11902281f9105&el=desc) will **increase** coverage by `0.00%`.
> The diff coverage is `100.00%`.

[...]

Issue #1061, opened 2021-04-03
Add flags similar to pydocstyle's --add-select/--add-ignore


asottile:
In GitLab by @sigmavirus24 on Aug 19, 2017, 04:49

These flags are quite useful for folks who merely want to extend the existing select and ignore lists. That said, we've found in pydocstyle that the `--add` verb is confusing to some. I would suggest `--extend-(select|ignore)` instead.

asottile:
In GitLab by @sigmavirus24 on Aug 19, 2017, 04:49

cc @mfcabrera

asottile:
In GitLab by @scop on May 13, 2018, 14:08

mentioned in merge request !233

asottile:
In GitLab by @fwinkl on Jun 8, 2020, 07:22

mentioned in merge request !437

MarcoGorelli:
Bit confused about this issue - doesn't flake8 already have extend-ignore?

sigmavirus24:
Yes but it doesn't have `--extend-select`

asottile:
via #1312

PR #1312, opened 2021-04-10
Add --extend-select command line argument

fwinkl:
Implement `--extend-select` command line argument following what was
done for `--extend-ignore` in !233 (see placeholder issue #1274). This option can be used to
selectively add individual codes without overriding the default list
entirely.

Addresses the remaining item of issue #1061.

### DIFF
--- a/docs/source/user/options.rst
+++ b/docs/source/user/options.rst
@@ -68,6 +68,8 @@ Index of Options
 
 - :option:`flake8 --select`
 
+- :option:`flake8 --extend-select`
+
 - :option:`flake8 --disable-noqa`
 
 - :option:`flake8 --show-source`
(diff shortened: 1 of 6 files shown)


docs/source/user/options.rst
```diff
@@ -632,6 +634,38 @@ Options and their Descriptions
             F
 
 
+.. option:: --extend-select=<errors>
+
+    :ref:`Go back to index <top>`
+
+    .. versionadded:: 3.9.0
```
asottile, line 641: the next version will be 4.0.0

src/flake8/main/options.py
```diff
@@ -272,6 +273,16 @@ def register_default_options(option_manager):
...
     )
 
+    add_option(
+        "--extend-select",
+        metavar="errors",
+        default="",
+        parse_from_config=True,
+        comma_separated_list=True,
+        help="Comma-separated list of errors and warnings to add to the list"
+        " of selected ones. For example, ``--extend-select=E4,E51,W234``.",
```
asottile, line 283: I would write this like:
```python
help=(
    "part one of the string "
    "part 2 of the string."
),
```

just so the words line up

asottile (commented):
otherwise seems fine

src/flake8/main/options.py
fwinkl: Since this is the style used for all other help messages, I would go with consistency here and not change it.

asottile: please fix this, it's one fewer for me to fix later

asottile (approved):

sigmavirus24:
Thanks @fwinkl !

PR #1317, opened 2021-04-15
Fix bug for plugins using extend_default_ignore

sigmavirus24:
Since Flake8 3.0 we've had the ability for plugins to use
`extend_default_ignore` to register codes they want disabled by default.
This, however, was a permanent disabling unfortunately. Our code didn't
have a way of understanding that this new set of `ignore` codes was
actually the 'default' set for that run. Much like the
extended_select_list, we now attach extended_ignore_list to be able to
confidently determine if the ignore we get in the DecisionEngine is
actually the Default Ignore list and what plugins what us to ignore by
default.

Refs https://github.com/PyCQA/pep8-naming/pull/157

--- 

Extra entertainment for Anthony
Screenshot-2021-04-15T09:16:52

### DIFF
--- a/src/flake8/options/aggregator.py
+++ b/src/flake8/options/aggregator.py
@@ -48,6 +48,10 @@ def aggregate_options(
     # Extend the default ignore value with the extended default ignore list,
     # registered by plugins.
     extended_default_ignore = manager.extended_default_ignore.copy()
+    # Let's store our extended default ignore for use by the decision engine
+    default_values.extended_default_ignore = (
+        manager.extended_default_ignore.copy()
+    )
     LOG.debug(
         "Extended default ignore list: %s", list(extended_default_ignore)
     )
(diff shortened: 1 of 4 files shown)


asottile (approved):
lg2m!  thanks for reformatting the tests, that was really bugging me :)

looks like a handful are over the length limit now, let me know if you want me to fix this for you

sigmavirus24:
Black reformatted for me. I'll fix the others too

sigmavirus24:
I'm not sure what your release plans are @asottile so I'll let you merge this

asottile:
I'm not planning to release for a while, but if this is important to you I can backport it to 3.9.x -- let me know!

sigmavirus24:
If you have docs around doing the backport I can also do it. I don't want to keep @ecolell hanging on that PR for too long. (Also knowing full well that pep8-naming will likely need a `>=3.9.x` for whatever value of `x` it ends up being.

asottile:
alrighty!  released this as 3.9.1

I decided to run the release myself since there were significant conflicts

sigmavirus24:
Thank you!

Issue #234, opened 2021-04-03
pyproject.toml (PEP 518) support


asottile:
In GitLab by @valignatev on Jun 11, 2018, 10:58

Hi! Is there a plan to support `pyproject.toml` from [PEP 518](https://www.python.org/dev/peps/pep-0518/)?
Specifically, it would be nice if flake8  could understand `[tool.flake8]` section in the `pyproject.toml` file.

asottile:
In GitLab by @jgirardet on Jul 28, 2018, 06:04

Hi,
I'm ok to do it.
I have a question for maintainers concerning the toml thing, which way would you like it to be implemented ?
some possibilities:
 - add a toml libray to flake8 dependency, which one (toml, tomlkit, other)
 - only use pyproject.toml  if a toml libray is importable (which would let to the user to add a toml lib in his requirements. still the question, which toml libray would we support ?
 - simply parse the pyproject.toml to extract the [tool.flake] section and deal with. not the best, probably not every usecase, but could be enough.

As simple implementation, all those possibility should return a dict to add to a configPArser.

Thank for your advice.
Jimmy

asottile:
In GitLab by @valignatev on Jul 28, 2018, 07:22

I'm nobody, of course (in the terms of flake8 project), but I personally think that `pyproject.toml` should be the first-class citizen since it's in the PEP. So the first option.

asottile:
In GitLab by @tmose1106 on Jul 29, 2018, 11:58

I am also not affiliate with flake8 in any way, but here is my two cents:

At the moment, is hard to tell which toml library to choose. As of today I have found the toml library to work just fine for parsing. It is also worth noting that if poetry picks up usage in the future, tomlkit could very well become the dominant toml library.

Specifically to @jgirardet's question, I feel that the second option would make the most sense at the moment. Simply make the [toml](https://github.com/uiri/toml) library an optional dependency and if there is a `pyproject.toml` file present, use it. I feel that the first option can be best handled as a future feature request once the dust has settled some more on poetry adoption so that we can make an informed choice on a mandatory toml library dependency and phase out some of the other configuration methods.

asottile:
In GitLab by @jgirardet on Jul 30, 2018, 03:47

mentioned in merge request !245

asottile:
In GitLab by @d_ on Jan 30, 2019, 02:38

mentioned in commit d_/dlc-gui@248a6b88d5d620dac793b7856fc217b0e1210e5d

[...]

PR #1332, opened 2021-05-11
Add support for reading config from pyproject.toml

eganjs:
With mypy support for `pyproject.toml` now merged (https://github.com/python/mypy/pull/10219), I think it would be good to try and add feature this to flake8 as well.

I have tried to implement this in a way to minimise pain, hopefully that's been achieved here.

Closes https://github.com/PyCQA/flake8/issues/234


asottile:
please read #234

PR #1333, opened 2021-05-15
Sort errors list in help

nvuillam:
[Mega-Linter](https://github.com/nvuillam/mega-linter) generates its help from flake8 --help

An automated job detects when there is a new version of help and triggers actions from this event
But as the list of returned errors is randomly ordered for every flake8 --help called, the result is never the same and there are false updates detection

Sorting the list of extensions will make help text to be always the same for the same version :)

### DIFF
--- a/src/flake8/main/options.py
+++ b/src/flake8/main/options.py
@@ -207,7 +207,7 @@ def register_default_options(option_manager):
     add_option(
         "--ignore",
         metavar="errors",
-        default=",".join(defaults.IGNORE),
+        default=",".join(sorted(defaults.IGNORE)),
         parse_from_config=True,
         comma_separated_list=True,
         help="Comma-separated list of errors and warnings to ignore (or skip)."


asottile:
that sequence is not randomly ordered, so your problem lies somewhere else:  https://github.com/PyCQA/flake8/blob/4fdb0a2b2fda7a20ca92b7fa8b5bf917fa74a6b5/src/flake8/defaults.py#L15

nvuillam:
@asottile any idea where I could find the origin ?

I assure you that in help output it is random

ex:

Nico@DESKTOP-E6MMMQR MINGW64 /c/Work/gitPerso/mega-linter-3 (bugs/swiftlint)
$ flake8 --help
usage: flake8 [options] file file ...

positional arguments:
  filename

optional arguments:
  -h, --help            show this help message and exit
  -v, --verbose         Print more information about what is happening in flake8. This option is repeatable and will increase verbosity each time it is   
                        repeated.
  --output-file OUTPUT_FILE
                        Redirect report to a file.
  --append-config APPEND_CONFIG
                        Provide extra config files to parse in addition to the files found by Flake8 by default. These files are the last ones read and   
                        so they take the highest precedence when multiple files provide the same option.
  --config CONFIG       Path to the config file that will be the authoritative config source. This will cause Flake8 to ignore all other configuration    
                        files.
  --isolated            Ignore all configuration files.
  --version             show program's version number and exit
  -q, --quiet           Report only file names, or nothing. This option is repeatable.
  --count               Print total number of errors and warnings to standard error and set the exit code to 1 if total is not empty.
  --diff                Report changes only within line number ranges in the unified diff provided on standard in by the user.
  --exclude patterns    Comma-separated list of files or directories to exclude. (Default: ['.svn', 'CVS', '.bzr', '.hg', '.git', '__pycache__', '.tox',  
                        '.eggs', '*.egg'])
  --extend-exclude patterns
                        Comma-separated list of files or directories to add to the list of excluded ones.
  --filename patterns   Only check for filenames matching the patterns in this comma-separated list. (Default: ['*.py'])
  --stdin-display-name STDIN_DISPLAY_NAME
                        The name used when reporting errors from code passed via stdin. This is useful for editors piping the file contents to flake8.    
                        (Default: stdin)
  --format format       Format errors according to the chosen formatter.
  --hang-closing        Hang closing bracket instead of matching indentation of opening bracket's line.
  --ignore errors       Comma-separated list of errors and warnings to ignore (or skip). For example, ``--ignore=E4,E51,W234``. (Default: **['E123','E226', 'E126', 'E704', 'W503', 'E24', 'W504', 'E121'])**
  --extend-ignore errors
                        Comma-separated list of errors and warnings to add to the list of ignored ones. For example, ``--extend-ignore=E4,E51,W234``.     
  --per-file-ignores PER_FILE_IGNORES
                        A pairing of filenames and violation codes that defines which violations to ignore in a particular file. The filenames can be     
                        specified in a manner similar to the ``--exclude`` option and the violations work similarly to the ``--ignore`` and ``--select``  
                        options.
  --max-line-length n   Maximum allowed line length for the entirety of this run. (Default: 79)
  --max-doc-length n    Maximum allowed doc line length for the entirety of this run. (Default: None)
  --select errors       Comma-separated list of errors and warnings to enable. For example, ``--select=E4,E51,W234``. (Default: ['E', 'F', 'W', 'C90'])   
  --disable-noqa        Disable the effect of "# noqa". This will report errors on lines with "# noqa" at the end.
  --show-source         Show the source generate each error or warning.
  --statistics          Count errors and warnings.
  --enable-extensions ENABLE_EXTENSIONS
                        Enable plugins and extensions that are otherwise disabled by default
  --exit-zero           Exit with status code "0" even if there are errors.
  --install-hook {git,mercurial}
                        Install a hook that is run prior to a commit for the supported version control system.
  -j JOBS, --jobs JOBS  Number of subprocesses to use to run checks in parallel. This is ignored on Windows. The default, "auto", will auto-detect the    
                        number of processors available to use. (Default: auto)
  --tee                 Write to stdout and output-file.
  --benchmark           Print benchmark information about this run of Flake8
  --bug-report          Print information necessary when preparing a bug report

mccabe:
  --max-complexity MAX_COMPLEXITY
                        McCabe complexity threshold

pyflakes:
  --builtins BUILTINS   define more built-ins, comma separated
  --doctests            also check syntax of the doctests
  --include-in-doctest INCLUDE_IN_DOCTEST
                        Run doctests only on these files
  --exclude-from-doctest EXCLUDE_FROM_DOCTEST
                        Skip these files when running doctests

Installed plugins: mccabe: 0.6.1, pycodestyle: 2.6.0, pyflakes: 2.2.0

Nico@DESKTOP-E6MMMQR MINGW64 /c/Work/gitPerso/mega-linter-3 (bugs/swiftlint)
$ flake8 --help
usage: flake8 [options] file file ...

positional arguments:
  filename

optional arguments:
  -h, --help            show this help message and exit
  -v, --verbose         Print more information about what is happening in flake8. This option is repeatable and will increase verbosity each time it is
                        repeated.
  --output-file OUTPUT_FILE
                        Redirect report to a file.
  --append-config APPEND_CONFIG
                        Provide extra config files to parse in addition to the files found by Flake8 by default. These files are the last ones read and
                        so they take the highest precedence when multiple files provide the same option.
  --config CONFIG       Path to the config file that will be the authoritative config source. This will cause Flake8 to ignore all other configuration 
                        files.
  --isolated            Ignore all configuration files.
  --version             show program's version number and exit
  -q, --quiet           Report only file names, or nothing. This option is repeatable.
  --count               Print total number of errors and warnings to standard error and set the exit code to 1 if total is not empty.
  --diff                Report changes only within line number ranges in the unified diff provided on standard in by the user.
  --exclude patterns    Comma-separated list of files or directories to exclude. (Default: ['.svn', 'CVS', '.bzr', '.hg', '.git', '__pycache__', '.tox',  
                        '.eggs', '*.egg'])
  --extend-exclude patterns
                        Comma-separated list of files or directories to add to the list of excluded ones.
  --filename patterns   Only check for filenames matching the patterns in this comma-separated list. (Default: ['*.py'])
  --stdin-display-name STDIN_DISPLAY_NAME
                        The name used when reporting errors from code passed via stdin. This is useful for editors piping the file contents to flake8.    
                        (Default: stdin)
  --format format       Format errors according to the chosen formatter.
  --hang-closing        Hang closing bracket instead of matching indentation of opening bracket's line.
  --ignore errors       Comma-separated list of errors and warnings to ignore (or skip). For example, ``--ignore=E4,E51,W234``. **(Default: ['W504', 'E126', 'E226', 'E121', 'E704', 'W503', 'E24', 'E123'])**
  --extend-ignore errors
                        Comma-separated list of errors and warnings to add to the list of ignored ones. For example, ``--extend-ignore=E4,E51,W234``.     
  --per-file-ignores PER_FILE_IGNORES
                        A pairing of filenames and violation codes that defines which violations to ignore in a particular file. The filenames can be     
                        specified in a manner similar to the ``--exclude`` option and the violations work similarly to the ``--ignore`` and ``--select``  
                        options.
  --max-line-length n   Maximum allowed line length for the entirety of this run. (Default: 79)
  --max-doc-length n    Maximum allowed doc line length for the entirety of this run. (Default: None)
  --select errors       Comma-separated list of errors and warnings to enable. For example, ``--select=E4,E51,W234``. (Default: ['E', 'F', 'W', 'C90'])   
  --disable-noqa        Disable the effect of "# noqa". This will report errors on lines with "# noqa" at the end.
  --show-source         Show the source generate each error or warning.
  --statistics          Count errors and warnings.
  --enable-extensions ENABLE_EXTENSIONS
                        Enable plugins and extensions that are otherwise disabled by default
  --exit-zero           Exit with status code "0" even if there are errors.
  --install-hook {git,mercurial}
                        Install a hook that is run prior to a commit for the supported version control system.
  -j JOBS, --jobs JOBS  Number of subprocesses to use to run checks in parallel. This is ignored on Windows. The default, "auto", will auto-detect the    
                        number of processors available to use. (Default: auto)
  --tee                 Write to stdout and output-file.
  --benchmark           Print benchmark information about this run of Flake8
  --bug-report          Print information necessary when preparing a bug report

mccabe:
  --max-complexity MAX_COMPLEXITY
                        McCabe complexity threshold

pyflakes:
  --builtins BUILTINS   define more built-ins, comma separated
  --doctests            also check syntax of the doctests
  --include-in-doctest INCLUDE_IN_DOCTEST
                        Run doctests only on these files
  --exclude-from-doctest EXCLUDE_FROM_DOCTEST
                        Skip these files when running doctests

Installed plugins: mccabe: 0.6.1, pycodestyle: 2.6.0, pyflakes: 2.2.0

asottile:
I'd suggest cloning the repository and poking around, maybe utilize the debugger

sigmavirus24:
To be further clear, https://github.com/PyCQA/flake8/blob/4fdb0a2b2fda7a20ca92b7fa8b5bf917fa74a6b5/src/flake8/main/options.py#L210 shows that what you're getting isn't from a very recent version of Flake8. We format the default very differently so we wouldn't be using a list representation as you've got there, but instead a string that should look like `E121,E123,E126,E226,E24,E704,W503,W504`. It would appear this is reported against an unsupported version of Flake8 - namely one that would appear to be much older than 3 years old: https://github.com/PyCQA/flake8/commit/c58a4662d8920cf70ea688edd9eaf9d783a856a7

PR #1341, opened 2021-05-25
Update default ignored errors

abstractlyZach:
This new list should now reflect the default errors that are ignored as defined here:
https://github.com/PyCQA/flake8/blob/3bcadce2a9d91f4bb6afa97b5d2d1a82d7a1e937/src/flake8/defaults.py#L15


asottile:
> And you should see something like:

this is not meant to be kept up to date

PR #1363, opened 2021-07-16
Replace if statements with if expressions

monosans:


sigmavirus24:
Thanks for thinking of us but we don't accept purely cosmetic changes

PR #1367, opened 2021-08-01
Update --indent-size help format string

JacobHayes:
Fixes this warning:
```
WARNING  flake8.options.manager:manager.py:186 option --indent-size: please update `help=` text to use %(default)s instead of %default -- this will be an error in the future
```

### DIFF
--- a/src/flake8/main/options.py
+++ b/src/flake8/main/options.py
@@ -260,7 +260,7 @@ def register_default_options(option_manager):
         metavar="n",
         default=defaults.INDENT_SIZE,
         parse_from_config=True,
-        help="Number of spaces used for indentation (Default: %default)",
+        help="Number of spaces used for indentation (Default: %(default)s)",
     )
 
     add_option(


src/flake8/main/options.py
```diff
@@ -260,7 +260,7 @@ def register_default_options(option_manager):
         metavar="n",
         default=defaults.INDENT_SIZE,
         parse_from_config=True,
-        help="Number of spaces used for indentation (Default: %default)",
+        help="Number of spaces used for indentation (Default: %(default))",
```
asottile, line 263: read the error message carefully!

JacobHayes: Oops, missing `s`

JacobHayes:
```suggestion
        help="Number of spaces used for indentation (Default: %(default)s)",
```

asottile (approved):

PR #1376, opened 2021-08-16
Fix type hint for build_ast

adamchainz:
`ast.parse` returns an `ast.Module` in its default mode, 'exec':

https://github.com/python/typeshed/blob/3ce5502675d3c23394b592f00234fbdfc743a7d5/stdlib/ast.pyi#L159-L167

Therefore this type hint can be more specific.

### DIFF
--- a/src/flake8/processor.py
+++ b/src/flake8/processor.py
@@ -218,7 +218,7 @@ class FileProcessor:
             (previous_row, previous_column) = end
         return comments, logical, mapping
 
-    def build_ast(self) -> ast.AST:
+    def build_ast(self) -> ast.Module:
         """Build an abstract syntax tree from the list of lines."""
         return ast.parse("".join(self.lines))
 


asottile:
I don't think this actually matters, flake8 passes this around as an opaque value and expects that plugins should accept any `AST`

adamchainz:
It may be that flake8 wants the flexibility to pass any `AST` type, but in practice it seems likely plugins depend on it being a `Module`. When adding type hints I found at least one of my plugins, `flake8-no-pep420`, assumed it is passed a `Module` rather than an `AST`.

asottile:
I'm leaning towards: your plugin should fix its assumptions instead of constraining flake8 (also it doesn't actually need `Module`, you can always use line 1)

adamchainz:
> your plugin should fix its assumptions instead of constraining flake8

Done - stopped looking at `tree` altogether.

> you can always use line 1

Doesn't that it break shebangs on some platforms to add a `# noqa` comment on the end?

asottile:
> 
> 
> > your plugin should fix its assumptions instead of constraining flake8
> 
> Done - stopped looking at `tree` altogether.
> 
> > you can always use line 1
> 
> Doesn't that it break shebangs on some platforms to add a `# noqa` comment on the end?

perhaps, though they can always use per-file-ignores instead 🤷

PR #1394, opened 2021-10-04
configuration.rst: Demonstrate how to set a boolean value

cclauss:
The current example config shows how to set string, list, and int values but does not show how to configure a Boolean value that requires `True` or `False` in a file-based config but not when configured on the command line.

### DIFF
--- a/docs/source/user/configuration.rst
+++ b/docs/source/user/configuration.rst
@@ -117,6 +117,7 @@ Let's actually look at |Flake8|'s own configuration section:
     ignore = D203
     exclude = .git,__pycache__,docs/source/conf.py,old,build,dist
     max-complexity = 10
+    show-source = True
 
 This is equivalent to:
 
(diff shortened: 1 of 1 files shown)


sigmavirus24 (commented):
I don't particularly feel like this is necessary as these are short examples of real-world usage. We document how to do this for _each_ option over on https://flake8.pycqa.org/en/latest/user/options.html

As an example: https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-show-source

sigmavirus24:
So to be clear,

1. This section is using our own configuration file as an example to demonstrate how the options are used and this option isn't set there
2. That said, it sounds like the options page isn't as discoverable? Would it help to link to that page from the configuration section to make it easier to find? Something like 

   ```reST
   Not all command-line options are configurable via a file. To know which are and which aren't, how to configure them, head over to :ref:`options` to get more details about the built-in CLI flags.
   ```

PR #1404, opened 2021-10-11
Drop support for Home and XDG config files

sigmavirus24:
This has been a huge support burden for us. I seriously considered doing
this in 3.0 but caved to a vocal minority and the desire to keep as much
backwards compatibility as possible. At this point, however, I'm done
witnessing the abuse Anthony has to suffer over this and I'm done with
the undue hostility that people who don't bother to read the docs
display. Hopefully, this eases that a bit.


willbelr:
What is the alternative? How to ignore specific error codes from now on?
I use it through Vim ale so I'm not sure how I could pass a cli argument.

sigmavirus24:
I use it through vim ale as well. I have no issues with per-project configuration.

PR #1423, opened 2021-10-13
remove 3.5 mention

dulmandakh:
Python 3.5 is no longer supported, but mentioned in the homepage


asottile:
is there a way to reword this to avoid a specific version being mentioned? otherwise this will need annoying updates forever

sigmavirus24:
It's intended to illustrate a point not be a mention of support. I'm with @asottile

Issue #1430, opened 2021-10-21
increase batches per worker, ~15% faster


collinanderson:
*Please describe how you installed Flake8*

```
$ sudo apt install flake8. # Ubuntu 20.04.3. I might pip-install if this feature request gets released.
```

*Please provide the exact, unmodified output of `flake8 --bug-report`*

```
{
  "dependencies": [
    {
      "dependency": "entrypoints",
      "version": "0.3"
    }
  ],
  "platform": {
    "python_implementation": "CPython",
    "python_version": "3.8.10",
    "system": "Linux"
  },
  "plugins": [
    {
      "is_local": false,
      "plugin": "mccabe",
      "version": "0.6.1"
    },
    {
      "is_local": false,
      "plugin": "pycodestyle",
      "version": "2.5.0"
    },
    {
      "is_local": false,
      "plugin": "pyflakes",
      "version": "2.1.1"
    }
  ],
  "version": "3.7.9"
}
```

*Please describe the problem or feature*

Hi All. Thanks so much for maintaining `flake8`.

Edit: ignore the sorting/optimizing chunks part. It seems to possibly have a negative benefit on large code-bases, though maybe still be _slightly_ helpful for smaller codebases, though probably isn't worth the complexity.

I think flake8 should increase batches per work and/or optimize chunks, possibly as settings.

I have 496 files in my project, many of them are small `django` migrations and `__init__.py` files, but there are a few really big ones (15482 tokens, ~1000 physical lines), which end up making one worker (out of 4) quit much later than the others, so the large chunks (2 per worker) aren't balanced.

In my case a normal `flake8` run on my project takes ~3.8 seconds (tested 10 runs / 10), if I edit the code to force pool `chunksize` to 1 or 2, I get ~3.3 seconds, a 13% improvement in my case.

Another thing could help would be to `stat` all the files and sort by file size, biggest first, then deal them out to each worker to help keep avoid a worker getting too much work at the end. For me when the files are cached in memory, it takes 2-4ms to stat and sort the files. That makes a larger `chunksize` more worth it:

```
        # in run_parallel(self):
        import os
        chunksize = calculate_pool_chunksize(len(self.checkers), self.jobs)
        # sort files by file size, largest first
        self.checkers.sort(key=lambda checker: os.stat(checker.filename).st_size, reverse=True)
        # deal the files out to the worker chunks
        new_sorted_checkers = []
        chunks = [[] for x in range(self.jobs)]
[...]

PR #1432, opened 2021-10-21
Increase batches per worker to 16, ~15% faster

collinanderson:
see issue #1430


asottile:
https://github.com/PyCQA/flake8/issues/1430#issuecomment-1293794436

PR #1438, opened 2021-11-04
Don't pin importlib-metadata

stephenfin:
In commit 975a3f45334861ebd8960c00e881443c23654bca, the `importlib-metadata` dependency was pinned to < 4.3. This was apparently to resolve some issues with that package that occured at that time. However, said issues are no longer present and this pinning is preventing packages that have `flake8` as a dependency from using the newer releases of `importlib-metadata` (in my case, this is preventing us from bumping `flake8` used by the OpenStack `hacking` package).

This PR proposes removing the cap as it's no longer necessary and is actively causing issues. I'm hoping that this can be included in a future patch release (i.e. 4.0.2) but if not, I guess a future minor release will have to suffice.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,7 +42,7 @@ install_requires =
     mccabe>=0.6.0,<0.7.0
     pycodestyle>=2.8.0,<2.9.0
     pyflakes>=2.4.0,<2.5.0
-    importlib-metadata<4.3;python_version<"3.8"
+    importlib-metadata; python_version<"3.8"
 python_requires = >=3.6
 
 [options.packages.find]


stephenfin:
fwiw, it's pretty self-explanatory but here's the error message I get when I try to bump the `flake8` dependency for `hacking`:

```
$ tox -e py36                                                                                                                                                                                                      
py36 create: /home/stephenfin/openstack/hacking/.tox/py36                                                                                                                                              
py36 installdeps: -chttps://releases.openstack.org/constraints/upper/master, -r/home/stephenfin/openstack/hacking/requirements.txt, -r/home/stephenfin/openstack/hacking/test-requirements.
txt                                                                                                                                                                                                                
ERROR: invocation failed (exit code 1), logfile: /home/stephenfin/openstack/hacking/.tox/py36/log/py36-1.log                                                                                           
==================================================================================================== log start ====================================================================================================
Collecting flake8~=4.0.0                                                                                                                                                                                           
  Using cached flake8-4.0.1-py2.py3-none-any.whl (64 kB)
...
Collecting flake8~=4.0.0
  Downloading flake8-4.0.0-py2.py3-none-any.whl (64 kB)
ERROR: Cannot install -r /home/stephenfin/openstack/hacking/requirements.txt (line 1) because these package versions have conflicting dependencies.

The conflict is caused by:
    flake8 4.0.1 depends on importlib-metadata<4.3; python_version < "3.8"
    flake8 4.0.0 depends on importlib-metadata<4.3; python_version < "3.8"
    The user requested (constraint) importlib-metadata===4.8.1

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
WARNING: You are using pip version 21.2.4; however, version 21.3.1 is available.
You should consider upgrading via the '/home/stephenfin/openstack/hacking/.tox/py36/bin/python -m pip install --upgrade pip' command.
```

We're seeing it because we use shared constraints files that are capping various dependencies, but obviously this would be an issue for a normal package that specified e.g. `importlib-metadata>4.3`.

sigmavirus24:
Hey @stephenfin I believe unpinning this will start emitting deprecation warnings for the API flake8 uses and nudging towards the unstable API intended to replace it. We're not intending to move to that API and the deprecation warnings are going to create maintenance burden for us. (A very small team.) Ignoring the harassment and character assassination we've already received from core python developers for not finding this to be the best path forward. I don't think we'll e lifting this pin anytime soon

stephenfin:
The commit I'm reverting here was very light on detail so I've no idea what I've walked into here :smile: All I know is that without this, we can't use flake8 > 4.x in OpenStack-land.

Have you any links that you could point me to providing more detail on this issue? It would be nice to find some way to resolve this that works for both the flake8 maintainers and the OpenStack community at large.

stephenfin:
Oh, okay, this is https://github.com/python/importlib_metadata/pull/319. Hmm...

sigmavirus24:
There are also discussions on discuss.python.org and there have been back channel discussions of Flake8 being unreasonable for taking someone at their word that an API was stable and should be relied upon while saying the new stable API could fly out the window in the next release

jeffnyman:
I see this PR is closed as are the tickets that reference the issue -- but what's the resolution for people who want to get work done on projects? When all this work is closed, guidance would be appreciated. (Although it sounds like that's what's being mentioned in the above referenced link as well: no guidance.)

Is the recourse to simply downgrade flake8 and pin it to version prior to 4.0.1?

alexpulver:
@jeffnyman there are a few suggestions in https://github.com/PyCQA/flake8/issues/1522

PR #1439, opened 2021-11-05
Add --required-plugins and --allowed-plugins

### LINKED ISSUE #283: Add a "require plugin" option
In GitLab by @mschwager on Jan 7, 2019, 07:34

Hi all,

I'd like to request an additional option to `flake8` for requiring certain plugin(s) are installed and selected during a run. For example,

```
$ flake8 --require=XYZ --select=XYZ /path/to/lint
```

I could also envision the new option implies a `--select`. Naming like `--require-plugins` also makes sense to me.

This would be very helpful for reminding developers to have a set list of `flake8` plugins installed on their machine. For example, a common use case may be that a repository uses a `Makefile` for various common operations including linting:

```
lint:
    flake8 --select=XYZ /path/to/lint

.PHONY: lint
```

Both the CI/CD system and developers use the `Makefile` to run these common operations, i.e. `make lint`. However, if the CI/CD system has the required plugins, but the developer's machine does not, then the local dev machine will silently ignore the selected rules:

```
$ flake8 --select=DOESNTEXIST /path/to/lint
$ echo $?
0
```

I'm envisioning a flag that would loudly complain if a required plugin is not installed:

```
lint:
    flake8 --require=ABC,XYZ /path/to/lint

.PHONY: lint
```

Keying off the plugin's prefix, similar to `--select` and `--ignore`, seems to be the most consistent solution in my opinion.

Another option could be for `flake8` to complain loudly if a plugin in a `--select` flag doesn't exist. However, a new flag makes more sense to me for ensuring backwards compatibility and because this seems like a net-new feature.

Thoughts?

### LINKED ISSUE #488: Having flake8 extensions install at system level will break projects and there is no way to disable them
In GitLab by @ssbarnea on Apr 30, 2016, 14:27

Here is a very simple to replicate use case, just install `flake8-docstrings` and now run flake8 on any project that was already using it. 

You will have a change of over 99% to fail passing the flake8 tests, and there is no way to disable those plugins from inside the `setup.cfg`.

I raise a bug month than a month ago on `flake8-docstrigs` regarding this issue and got no reply at all, it seems that the project not maintained at all.

Assuming that that's not an isolated case we need a way to assure that the linting settings are locked within the `setup.cfg`, so the final result will not depend on whichever additional modules you may have installed on your system-modules.

I know that someone with inevitably try to propose a virtual environment with `--no-site-packages` but please remember that those are not always possible, there are several use-cases where people are forced to import system-packages (like on continous integration systems where you cannot afford to increase the footprint of each build with 500MB just because your projects is using numpy and pandas).

Flake8 needs an option to tell it to use a specific list of plugins and to ignore others.

See https://gitlab.com/pycqa/flake8-docstrings/issues/10

sigmavirus24:
Closes #283
Closes #488

### DIFF
--- a/src/flake8/api/legacy.py
+++ b/src/flake8/api/legacy.py
@@ -38,7 +38,7 @@ def get_style_guide(**kwargs):
         ignore_config_files=prelim_opts.isolated,
     )
 
-    application.find_plugins(config_finder)
+    application.find_plugins(config_finder, prelim_opts)
     application.register_plugin_options()
     application.parse_configuration_and_cli(
         config_finder,
--- a/src/flake8/exceptions.py
+++ b/src/flake8/exceptions.py
@@ -1,5 +1,6 @@
 """Exception classes for all of Flake8."""
 from typing import Dict
+from typing import List
 
 
 class Flake8Exception(Exception):
@@ -68,4 +69,25 @@ class PluginExecutionFailed(Flake8Exception):
         return self.FORMAT % {
             "name": self.plugin["plugin_name"],
             "exc": self.original_exception,
+        }
+
+
+class PluginMissingError(Flake8Exception):
+    """A plugin that was required was not found."""
+
+    FORMAT = "User required %(plugins)s but %(missing)s was not found."
+
+    def __init__(
+        self, required_plugins: List[str], missing_plugins: List[str]
+    ) -> None:
+        """Store the information passed in to format the exception message."""
+        self.required_plugins = required_plugins
+        self.missing_plugins = missing_plugins
+        super().__init__(required_plugins, missing_plugins)
+
+    def __str__(self) -> str:
+        """Format our exception message."""
+        return self.FORMAT % {
+            "plugins": ", ".join(self.required_plugins),
+            "missing": ", ".join(self.missing_plugins),
         }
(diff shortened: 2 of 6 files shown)


asottile:
I'm in the middle of reworking plugin loading and this goes in the other direction of what I'm doing

essentially the planned pipeline is:

1. initial opts
2. load all plugins
3. parse finalized opts
4. (then filtering/requiring plugins would land here)
5. then the rest of the execution pipeline

sigmavirus24:
> load all plugins

Loading them will break things and defeat the whole purpose of adding the options in the first place unless we're defining "load" differently.

asottile:
> 
> 
> > load all plugins
> 
> Loading them will break things and defeat the whole purpose of adding the options in the first place unless we're defining "load" differently.

import and parse their options -- I realize there's still situations where that breaks as well but I think in those cases it's expected that the user would uninstall or isolate using a virtualenv.  whereas I expect these options to be predominantly used to prevent the plugins from executing against files

sihil:
Hello. This is an awesome feature, thank you all. Any idea when you'll cut a release that includes these options? Confusingly, they are on the latest documentation but not in `4.0.1` which seems to be the latest available unless I'm mistaken. Thank you!

PR #1441, opened 2021-11-06
deprecate the --diff option

### LINKED ISSUE #1389: Current implementation of `--diff` is kind of broken by design
`--diff` currently makes assumptions that code changes will introduce lint errors *at the position of the code change* which may have been true for `pycodestyle` which only concerns itself with whitespace / style issues -- but it is not true for pyflakes or generally true for flake8 plugins

here is a small concrete example:

```diff
 import os
 
 def main():
-    print(os.getcwd())
+    print('hi')
```

`flake8 t.py` reports F401 on line 1, but `flake8 --diff` silently passes (even though the diff _caused_ that new issue)

as such, I think the option's behaviour needs to be changed in an incompatible way, or removed entirely

if kept, I think the only reasonable behaviour would be to limit changes to the file which a diff touched -- rather than the lines which were touched

asottile:
resolves #1389

### DIFF
--- a/docs/source/user/options.rst
+++ b/docs/source/user/options.rst
@@ -238,6 +238,11 @@ Options and their Descriptions
 
     :ref:`Go back to index <top>`
 
+    .. warning::
+
+        Due to hiding potential errors, this option is deprecated and will be
+        removed in a future version.
+
     Use the unified diff provided on standard in to only check the modified
     files and report errors included in the diff.
 
--- a/src/flake8/main/application.py
+++ b/src/flake8/main/application.py
@@ -191,6 +191,10 @@ class Application:
 
         self.running_against_diff = self.options.diff
         if self.running_against_diff:
+            LOG.warning(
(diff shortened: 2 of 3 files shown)


noamraph:
@asottile I would vote to un-deprecate this feature. I'm trying to improve the code style in my company, and as a first step I want to forbid commits that add new warnings. I hope that in the future I will be able to fix all style issues, but that will take a (long) while. Using the --diff option seems like the only way to prevent new style violations from being committed.

I understand that --diff is not perfect, as new warnings can be triggered. However, as it's a very useful feature, I suggest to replace the deprecation with a note explaining why new warnings might be introduced even if `flake8 --diff` didn't show anything.

asottile:
it's fundamentally broken so it is being removed

I recommend you either:
- use a code formatter to fix the existing issues
- work file-by-file to fix issues

more on that: https://github.com/readme/guides/code-review

PR #1457, opened 2021-11-14
improve coverage a bit

asottile:


src/flake8/checker.py
```diff
@@ -15,11 +16,6 @@
 from flake8 import processor
 from flake8 import utils
 
-try:
-    import multiprocessing.pool
-except ImportError:
-    multiprocessing = None  # type: ignore
-
```
asottile, line 22: left over from python 2.x days

PR #1463, opened 2021-11-15
split out file discovery and test it

asottile:


tox.ini
```diff
@@ -122,4 +122,5 @@ commands =
 extend-ignore = E203
 per-file-ignores =
     src/flake8/formatting/_windows_color.py: N806
+    tests/*: D
```
asottile, line 125: got tired of:

```python
def test_that_x_does_y():
    """Verify that X does Y."""
```

Issue #199, opened 2021-04-03
Config uses tox file even when no flake8 config in it


asottile:
In GitLab by @adamchainz on Nov 21, 2016, 03:51

*Please describe how you installed Flake8*

```
$ pip install flake8
```

*Please provide the exact, unmodified output of `flake8 --bug-report`*

```
$ flake8 --bug-report
{
  "dependencies": [
    {
      "dependency": "setuptools",
      "version": "23.1.0"
    }
  ],
  "platform": {
    "python_implementation": "CPython",
    "python_version": "2.7.12",
    "system": "Darwin"
  },
  "plugins": [
    {
      "plugin": "flake8-tidy-imports",
      "version": "1.0.0"
    },
    {
      "plugin": "mccabe",
      "version": "0.5.2"
    },
    {
      "plugin": "pycodestyle",
      "version": "2.2.0"
    },
    {
      "plugin": "pyflakes",
      "version": "1.3.0"
    }
  ],
  "version": "3.2.0"
}
```

*Please describe the problem or feature*

I have a project with many subprojects each using `tox`. I'd like to define my lint rules at the top level in a `setup.cfg` and have individual `tox.ini` files for each subproject.

Currently the existence of `tox.ini`, without a `flake8` header, causes `flake8` to consider itself 'configured' with just the defaults, and the search for the config stops. I would expect it to find the flake8 config in my `setup.cfg` instead.

It's also non-deterministic based upon the current directory when `flake8` is run.

My current workaround is to always invoke `flake8` with `--config` pointed to the absolute path of the config file, which isn't ideal.

*If this is a bug report, please explain with examples (and example code) what you expected to happen and what actually happened.*

```
$ tree .
.
├── setup.cfg
└── subproj
    ├── test.py
    └── tox.ini

1 directory, 3 files
$ cat setup.cfg
[flake8]
max-line-length = 120
$ cat subproj/tox.ini
[tox]
something = something
$ cat subproj/test.py
# this line is longer than 80 chars but less than 120 chars and so should be allowed, but it isn't
$ flake8  # passes
$ flake8 subproj/test.py  # fails
subproj/test.py:1:80: E501 line too long (98 > 79 characters)
$ cd subproj
$ flake8  # fails
./test.py:1:80: E501 line too long (98 > 79 characters)
```

asottile:
In GitLab by @OmeGak on Dec 28, 2017, 10:50

[...]

PR #1472, opened 2021-11-23
refactor and simplify configuration loading

asottile:


515k4:
I did some testing of released 4.0.1 and version with this merge and they do not behave the same. Moreover, in both cases, they could read your `~/.flake8`. Consider this example:

```
# Test setup
~$ echo -e "[flake8]\nmax-line-length = 88" > ~/.flake8
~$ mkdir project1 project2 project3
~$ echo 'print("This is very long line. This is very long line. This is very long line. End")' > project1/example.py
~$ cp project1/example.py project2
~$ cp project1/example.py project3
~$ touch project2/setup.cfg
~$ echo "[flake8]" > project3/setup.cfg
```

```
# Run with 4.0.1
~/project1$ flake8 example.py
~/project1$ cd ../project2
~/project2$ flake8 example.py
E501 line too long (84 > 79 characters)
~/project2$ cd ../project3
~/project3$ flake8 example.py
E501 line too long (84 > 79 characters)
```

```
# Run with freshly cloned repo (7a02d04)
~/project1$ flake8 example.py
~/project1$ cd ../project2
~/project2$ flake8 example.py
~/project2$ cd ../project3
~/project3$ flake8 example.py
example.py:1:80: E501 line too long (84 > 79 characters)
```

The documentation says: the configuration is in your project in one of `setup.cfg`, `tox.ini`, or `.flake8` (It doesn't say in this order but the order is also important). In 4.0.1 the project's directory is a directory with any of those files regarding its content. After refactoring, project's directory means some of those files must have `[flake8]` section. In any case, if no directory seems to be project's directory, it can accidently parse your ~/.flake8, which I understand it should not parse anymore.

I am fine with both of  those behaviors but I really suggest either more verbose documentation or flake8 itself. Now it is quite confusing and when I run flake8 on very simple files without project (like example or testing files) I get unexpected results.

asottile:
yes, this fixes a bug in the current versions where a configuration without `[flake8]` is considered a flake8 configuration

asottile:
specifically #199

515k4:
Good! But if that was a bug and I am not really using `[flake8]` in any of my `setup.cfg` nor `tox.ini`, my `~/.flake8` will continue to work. Is this intended?

asottile:
it's not supported but it may work by accident -- I would not rely on that behaviour

515k4:
I fear this is not the type of accident you want have in such a pedantic tool. You state in 4.0.0 changelog that _files will not be searched for in the user’s home directory (e.g., ~/.flake8)_ and suddenly the tool is searching exactly for that file because it's unclear what qualifies a directory to be a projects's directory.

Please consider adding a warning when using `~/.flake8` to not rely on it  and also consider updating the documentation. The file order and what is project directory exactly.

asottile:
I don't appreciate the attitude

PR #1480, opened 2021-11-28
More descriptive documentation about config files.

515k4:
Please consider this change in Configuration section in your documentation. I got quite confused with the section regarding recent changes in #1404 and #1472 and even #199 and unfortunatelly ​and unintentionally have been mean to your developer in trying to fix it. I am sorry and here is my take to be constructive.

### DIFF
--- a/docs/source/user/configuration.rst
+++ b/docs/source/user/configuration.rst
@@ -25,7 +25,8 @@ Configuration Locations
 =======================
 
 |Flake8| supports storing its configuration in your project in one of
-``setup.cfg``, ``tox.ini``, or ``.flake8``.
+``setup.cfg``, ``tox.ini``, or ``.flake8``. The configuration is not merged
+across those files and files are considered in that particular order.
 
 Values set at the command line have highest priority, then those in the
 project configuration file, and finally there are the defaults. However,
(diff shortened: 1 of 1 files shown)


docs/source/user/configuration.rst
```diff
@@ -25,7 +25,8 @@ Configuration Locations
 =======================
 
 |Flake8| supports storing its configuration in your project in one of
```
asottile, line 27: it already says "one of"

docs/source/user/configuration.rst
```diff
@@ -72,6 +73,15 @@ to be in the ``flake8`` section, which means it needs to start like so:
...
 
+.. note::
+
+   If your project does not use ``setup.cfg``, ``tox.ini``, neither
+   ``.flake8``, or they don't contain section ``flake8`` at all, the searching
+   for suitable configuration file will continue in project's parent
+   directories. This may lead to parsing no longer supported user-level
+   configuration at ``~/.flake8``. Please do not rely on this behaviour hence
+   we are supporting only per-project configuration.
+
```
asottile, line 84: I'd rather not document an unsupported behaviour

docs/source/user/configuration.rst
515k4: It is correct, it's just a little ambiguous. My first thought after reading original sentence was which one will be used if there is part of configuration in all of the files. I end up looking for answer in source code and here on github.

docs/source/user/configuration.rst
515k4: So just leave out the last two sentences? The beginning part is supported behaviour, isn't it?

PR #1481, opened 2021-11-29
allow additional default configuration locations

czandee:
Hi,
The current configuration finder only looks in the current directory and its parents. This makes it impossible to have a configuration in other default locations such as the userhome (e.g. ~/.flake8 or ~/.config/flake8)). The only way to address such configuration is to always use `--append-config` or `--config` options.
As it is quite common (I think) to use the same configuration for each project it would be nice if there are some default locations outside the current or project working directory, so flake8 can use these outside locations as fallback before using the internal defaults.
With this pull request I try to not change the original behaviour at all - only when flake8 does not find a suitable configuration my proposal is to scan some more locations before using the internal defaults (right now: ~/.config and ~).
Best regards, 
Christian

### DIFF
--- a/tests/unit/test_file_checker.py
+++ b/tests/unit/test_file_checker.py
@@ -39,7 +39,6 @@ def test_raises_exception_on_failed_plugin(tmp_path, default_options):
         "parameters": dict(),
         "plugin": mock.MagicMock(side_effect=ValueError),
     }
-    """Verify a failing plugin results in an plugin error"""
     fchecker = checker.FileChecker(
         str(foobar), checks=[], options=default_options
     )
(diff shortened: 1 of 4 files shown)


tests/unit/test_file_checker.py
```diff
@@ -39,7 +39,6 @@ def test_raises_exception_on_failed_plugin(tmp_path, default_options):
         "parameters": dict(),
         "plugin": mock.MagicMock(side_effect=ValueError),
     }
-    """Verify a failing plugin results in an plugin error"""
```
czandee, line 42: Frankly, I don't know why this `tox -e linters` error appeared. Has nothing to do with the PR itself.

sigmavirus24:
We specifically removed user configuration files

515k4:
After #1472 the `~/.flake8` will continue to work as a side effect of refactoring.

asottile:
@515k4 please stop

sigmavirus24:
@515k4 you've been blocked from contributing to the PyCQA for a month. If you continue this behaviour after that ban has expired, you'll be permanently banned

PR #1498, opened 2021-12-24
return an error if the explicitly specified config is missing

### LINKED ISSUE #1497: A missing explicitly specified config file is silently ignored
*Please describe how you installed Flake8*

```console
$ virtualenv venv
$ . venv/bin/activate
$ pip install flake8
```

*Please provide the exact, unmodified output of `flake8 --bug-report`*

```cnosole
$ flake8 --bug-report
{
  "dependencies": [],
  "platform": {
    "python_version": "3.8.8",
    "system": "Linux"
  },
  "plugins": [
    {
      "is_local": false,
      "plugin": "mccabe",
      "version": "0.6.1"
    },
    {
      "is_local": false,
      "plugin": "pycodestyle",
      "version": "2.8.0"
    },
    {
      "is_local": false,
      "plugin": "pyflakes",
      "version": "2.4.0"
    }
  ],
  "version": "4.0.1"
}
```

*Please describe the problem or feature*

flake8 silently ignores the specified config file if it is missing

```console
$ rm -f foo
$ touch t.py
$ flake8 --config foo t.py
$ echo $?
0
```

*If this is a bug report, please explain with examples (and example code) what you expected to happen and what actually happened.*

An error should be raised if the specified config file is missing.

Looks like the culprit is `configparser.read()`: ["If a file named in filenames cannot be opened, that file will be ignored"](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read)

rkm:
fixes #1497

### DIFF
--- a/src/flake8/options/config.py
+++ b/src/flake8/options/config.py
@@ -8,6 +8,7 @@ from typing import List
 from typing import Optional
 from typing import Tuple
 
+from flake8 import exceptions
 from flake8.options.manager import OptionManager
 
 LOG = logging.getLogger(__name__)
@@ -60,7 +61,10 @@ def load_config(
 
     cfg = configparser.RawConfigParser()
     if config is not None:
-        cfg.read(config)
+        if not cfg.read(config):
+            raise exceptions.ExecutionError(
+                f"The specified config file does not exist: {config}"
+            )
         cfg_dir = os.path.dirname(config)
     else:
         cfg_dir = pwd
--- a/tests/integration/test_main.py
+++ b/tests/integration/test_main.py
@@ -386,3 +386,21 @@ def test_early_keyboard_interrupt_does_not_crash(capsys):
     out, err = capsys.readouterr()
     assert out == "... stopped\n"
     assert err == ""
+
+
+def test_config_file_not_found(tmpdir, capsys):
+    """Ensure that an explicitly specified config file which is not found is an
+    error"""
+
+    expected = """\
+There was a critical error during execution of Flake8:
+The specified config file does not exist
+"""
+
+    with tmpdir.as_cwd():
+        tmpdir.join("t.py").write("print('hello hello world')\n")
+        assert cli.main(["--config", "missing.cfg", "t.py"]) == 1
+
+    out, err = capsys.readouterr()
+    assert out == expected
+    assert err == ""
--- a/tests/integration/test_main.py
+++ b/tests/integration/test_main.py
@@ -394,7 +394,7 @@ def test_config_file_not_found(tmpdir, capsys):
 
     expected = """\
 There was a critical error during execution of Flake8:
-The specified config file does not exist
+The specified config file does not exist: missing.cfg
 """
 
     with tmpdir.as_cwd():
--- a/tests/unit/test_options_config.py
+++ b/tests/unit/test_options_config.py
@@ -2,6 +2,7 @@ import configparser
 
 import pytest
 
+from flake8 import exceptions
 from flake8.main.options import register_default_options
 from flake8.options import config
 from flake8.options.manager import OptionManager
@@ -164,3 +165,8 @@ def test_parse_config_ignores_unknowns(tmp_path, opt_manager, caplog):
             'Option "wat" is not registered. Ignoring.',
         )
     ]
+
+
+def test_load_config_missing_file_raises_exception(capsys):
+    with pytest.raises(exceptions.ExecutionError):
+        config.load_config("foo.cfg", [])


src/flake8/main/application.py
```diff
@@ -115,8 +115,7 @@ def parse_preliminary_options(
 
     def exit_code(self) -> int:
         """Return the program exit code."""
-        assert self.options is not None
-        if self.options.exit_zero:
```
sigmavirus24, line 119: This is unrelated to your change. Please revert it

src/flake8/options/config.py
```diff
@@ -60,6 +61,10 @@ def load_config(
 
     cfg = configparser.RawConfigParser()
     if config is not None:
+        if not os.path.isfile(config):
+            raise exceptions.ExecutionError(
+                "The specified config file does not exist"
```
sigmavirus24, line 66: Why wouldn't you include the config value in the exception?

sigmavirus24:
We have not documented that we would raise an error if the configuration file doesn't exist and if we wanted to split hairs, this is backwards incompatible behaviour meaning we couldn't easily just accept this at the moment

src/flake8/options/config.py
rkm: have added this now

src/flake8/main/application.py
rkm: Without changing the two assertions, the output is:
```console
$ flake8 --config foo.cfg t.py
There was a critical error during execution of Flake8:
The specified config file does not exist: foo.cfg
  File "/home/rkm/dev/pycqa/flake8/venv/bin/flake8", line 33, in <module>
    sys.exit(load_entry_point('flake8', 'console_scripts', 'flake8')())
  File "/home/rkm/dev/pycqa/flake8/src/flake8/main/cli.py", line 22, in main
    app.run(argv)
  File "/home/rkm/dev/pycqa/flake8/src/flake8/main/application.py", line 395, in run
    assert self.options is not None
AssertionError
```
Compared to simply:
```console
$ flake8 --config foo.cfg t.py
There was a critical error during execution of Flake8:
The specified config file does not exist: foo.cfg
```
To me the traceback seems misleading, since it doesn't point to the actual call site. If that's acceptable then I can re-add the assertions, otherwise let me know if you have a better solution in mind.

rkm:
Thanks for reviewing. I hadn't considered there would be any worries about backwards incompatibility. Feel free to close this if you think that's a blocker. If so then could we at least update the docs to indicate the existing behaviour?

src/flake8/main/application.py
asottile: I think this should be

```python
    def exit_code(self) -> int:
        """Return the program exit code."""
        if self.catastrophic_failure:
            return 1
        assert self.options is not None
        if self.options.exit_zero:
            return 0
        else:
            return int(self.result_count > 0) 
```

eventually I want to factor away the `Application` class since the lateinit nullable attributes pose a lot of problems and we have to side-step the typechecker a lot

asottile: actually it's slightly more involved -- here's a fix for that:

```diff
diff --git a/src/flake8/main/application.py b/src/flake8/main/application.py
index ee46f6a..1894881 100644
--- a/src/flake8/main/application.py
+++ b/src/flake8/main/application.py
@@ -115,10 +115,13 @@ class Application:
 
     def exit_code(self) -> int:
         """Return the program exit code."""
-        if self.options and self.options.exit_zero:
-            return int(self.catastrophic_failure)
+        if self.catastrophic_failure:
+            return 1
+        assert self.options is not None
+        if self.options.exit_zero:
+            return 0
         else:
-            return int((self.result_count > 0) or self.catastrophic_failure)
+            return int(self.result_count > 0)
 
     def find_plugins(
         self,
@@ -390,6 +393,7 @@ class Application:
         except exceptions.EarlyQuit:
             self.catastrophic_failure = True
             print("... stopped while processing files")
-
-        if self.options and self.options.count:
-            print(self.result_count)
+        else:
+            assert self.options is not None
+            if self.options.count:
+                print(self.result_count)
```

actually, I think this is broken on `main` so I'll split it out to a separate patch

src/flake8/options/config.py
```diff
@@ -60,6 +61,10 @@ def load_config(
 
     cfg = configparser.RawConfigParser()
     if config is not None:
+        if not os.path.isfile(config):
+            raise exceptions.ExecutionError(
+                f"The specified config file does not exist: {config}"
+            )
         cfg.read(config)
```
asottile, line 68: a ~slightly more succinct way to write this is:

```python
if not cfg.read(config):
    raise exceptions.ExecutionError(...)
```

asottile (commented):
while this ~technically breaks backward compatibility -- I think this is a move in the right direction and I'm ok with this being a new error

just needs a slight rebase though 👍

asottile (approved):

PR #1529, opened 2022-01-19
update documentation links in issue templates

paulhfischer:


sigmavirus24 (approved):

.github/ISSUE_TEMPLATE/02_feature.yml
```diff
@@ -6,7 +6,7 @@ body:
       value: >
         Please read this brief portion of documentation before going any
         further:
-        http://flake8.pycqa.org/en/latest/internal/contributing.html#filing-a-bug
```
asottile, line 9: ah we still want them to read everything here because it requests that they search the issue tracker

sigmavirus24: My bad. I missed the anchor change

paulhfischer: okay my bad -- i thought the anchor was wrong as there is an extra paragraph for feature requests but i get your point @asottile.

PR #1544, opened 2022-01-24
remove slow debug() log

asottile:
flake8 spends ~5-6% of `flake8 -j1 src` on this line

___

personally I don't use any of the `debug(...)` logging information and prefer to use `pdb` myself.  I wouldn't be opposed to removing the rest of the debug lines as well -- though this one appears to be the most performance impacting one (even at the default log level)

### DIFF
--- a/src/flake8/checker.py
+++ b/src/flake8/checker.py
@@ -337,7 +337,6 @@ class FileChecker:
 
     def run_check(self, plugin: LoadedPlugin, **arguments: Any) -> Any:
         """Run the check in a single plugin."""
-        LOG.debug("Running %r with %r", plugin, arguments)
         assert self.processor is not None
         try:
             params = self.processor.keyword_arguments_for(


Pierre-Sassoulas:
I think I remember a conversation regarding this line in the gitlab repository but I can't find it anymore. The argument to keep was that the log was useful when asking an user for troubleshooting.

sigmavirus24:
I believe I added this because of how often plugin developers emailed me asking me why they were getting unexpected values passed in. My idea was this would help with development for them and help them support their plugins better. I'm open to a different set of instructions and tools for that. I'm not attached to this at all, but I don't want to lose that self-service capability for developers

sigmavirus24:
> alright I'm going to merge this for now -- I think perhaps I could add a noop `off_by_default` "debug" extension plugin which logs its inputs that could replace this.  the user could opt into this information via `flake8 --enable-extension=DBG1,DBG2,DBG3` -- though my gut is we should encourage plugin developers do do their own debugging via `breakpoint()` or `print(...)` *within their plugins*

That is in fact what I've encouraged in the past but occasionally they're trying to debug something from a user and this was hopefully a helper for them

Issue #751, opened 2021-04-03
Performance suggestion: do not run unselected plugins/checks


asottile:
In GitLab by @hugovk on Jun 5, 2020, 01:45

Please read this brief portion of documentation before going any further: http://flake8.pycqa.org/en/latest/internal/contributing.html#filing-a-bug

*Please describe how you installed Flake8*

```
$ pip install -U flake8
$ brew install flake8
# etc.
```

*Please provide the exact, unmodified output of `flake8 --bug-report`*

```
{
  "dependencies": [],
  "platform": {
    "python_implementation": "CPython",
    "python_version": "3.8.3",
    "system": "Darwin"
  },
  "plugins": [
    {
      "is_local": false,
      "plugin": "flake8_2020",
      "version": "1.6.0"
    },
    {
      "is_local": false,
      "plugin": "mccabe",
      "version": "0.6.1"
    },
    {
      "is_local": false,
      "plugin": "pycodestyle",
      "version": "2.6.0"
    },
    {
      "is_local": false,
      "plugin": "pyflakes",
      "version": "2.2.0"
    }
  ],
  "version": "3.8.2"
}
```

*Please describe the problem or feature*

I noticed that Flake8 takes the same time to run with `--select` as without. As shown using `-vv` verbosity, it runs all the plugins and checks regardless of `--select`, and only reports the selected ones afterwards.

Flake8 can sometimes take a long time to run on large codebases, and if it was possible to only run the selected checks, that would save a lot of time, CPU and power.

Would it be possible to only run selected checks/plugins? Rather than running them anyway and discarding that work when reporting?

---

# Docs

For reference, my emphasis.

`flake8 --help` says `--select` is for which ones to _enable_:

```
  --select errors       Comma-separated list of errors and warnings to enable. For example, ``--select=E4,E51,W234``.
                        (Default: ['E', 'F', 'W', 'C90'])
```

The docs are a bit more explicit:

> Specify the list of error codes you wish Flake8 to _report_.

 https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-select

---

# Example

An example running on the TensorFlow codebase:

```console
$ time flake8
...
flake8  323.91s user 4.31s system 98% cpu 5:32.78 total
$ time flake8 --select YTT
...
flake8 --select YTT  318.62s user 3.80s system 99% cpu 5:25.51 total
```

Both about the same, around 5m20s.

[...]

PR #1545, opened 2022-01-25
pregenerate the pycodestyle plugin to avoid call overhead

asottile:
this saves 10-12% off the execution time of `flake8 -j1 src/`

- best of 10 before: 1.469s
- best of 10 after: 1.301s

the implementation is a tiny bit fragile, though not exceedingly so as we pin `pycodestyle`.


andersk:
Would it be even better to generate that code and `exec()` it at runtime? That could give you a way to remove the overhead for deselected plugins (#751).

asottile:
> Would it be even better to generate that code and `exec()` it at runtime? That could give you a way to remove the overhead for deselected plugins (#751).

#751 is basically impossible without refactoring how pycodestyle works (checks in pycodestyle do not all fall under unique prefixes and criss cross over)

the codegen is also relatively slow so doing it at runtime would eliminate most of the performance gain

sigmavirus24 (approved):
I'm never really a fan of code generation but the improvement is hard to deny. I _suspect_ most of the speed comes from fewer `entry_points` given how abysmal that's always been (regardless of how much faster `importlib` is than `setuptools`/`pkg_resources`).

asottile:
surprisingly, most of the speedup comes from not needing to call `keywords_for`  and building small kwargs dictionaries many many times

PR #1548, opened 2022-01-28
Remove duplicated example from `options.rst`

sobolevn:
Here's how it looks right now: https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-enable-extensions

Снимок экрана 2022-01-28 в 11 40 42

Probably, this is not intended.


sobolevn:
Oh, I see `-` vs `_`

PR #1551, opened 2022-02-06
re-show pycodestyle in help after plugin gen

asottile:
accidentally regressed in #1545

### DIFF
--- a/src/flake8/plugins/finder.py
+++ b/src/flake8/plugins/finder.py
@@ -162,12 +162,11 @@ def _flake8_plugins(
 
         if ep.name == "F":
             yield Plugin(pyflakes_meta["name"], pyflakes_meta["version"], ep)
-        elif ep.name.startswith("pycodestyle"):
+        elif ep.name in "EW":
             # pycodestyle provides both `E` and `W` -- but our default select
             # handles those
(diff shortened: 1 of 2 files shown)


src/flake8/plugins/finder.py
```diff
@@ -162,12 +162,11 @@ def _flake8_plugins(
 
         if ep.name == "F":
             yield Plugin(pyflakes_meta["name"], pyflakes_meta["version"], ep)
-        elif ep.name.startswith("pycodestyle"):
+        elif ep.name in "EW":
             # pycodestyle provides both `E` and `W` -- but our default select
             # handles those
             # ideally pycodestyle's plugin entrypoints would exactly represent
             # the codes they produce...
```
sigmavirus24, line 169: This comment can probably disappear now?

asottile: it's still unfortunately accurate :( -- the `E` and `W` don't 1:1 match up to the codes that are provided

I eventually want at some point for those to be accurate so I can introduce:
- a helpful warning for plugin authors when their errors aren't showing up as expected (mismatch of entrypoint name and error codes)
- a way to skip plugins based on whether they'd be ignored

PR #1552, opened 2022-02-06
make --ignore order consistent

asottile:
for #1550 

I think this is worth doing anyway -- though we should reconsider whether showing the configured values as "default" in `--help` is desirable or not


sigmavirus24:
I believe I used a set because someone complained about having a long list of plugins which ended up including duplicates and the performance hit of checking the same code more than once. I was trying to deduplicate things without having to worry about checking if it was already present or keep 2 copies of things. :shrug:

PR #1556, opened 2022-02-09
Clarify that --count writes to standard output

### LINKED ISSUE #1554: Help text for `--count` option incorrectly specifies that the count is output to stderr

```console
$ pip install flake8==4.0.1
```

```json
{
  "dependencies": [],
  "platform": {
    "python_version": "3.9.7",
    "system": "Linux"
  },
  "plugins": [
    {
      "is_local": false,
      "plugin": "mccabe",
      "version": "0.6.1"
    },
    {
      "is_local": false,
      "plugin": "pycodestyle",
      "version": "2.8.0"
    },
    {
      "is_local": false,
      "plugin": "pyflakes",
      "version": "2.4.0"
    }
  ],
  "version": "4.0.1"
}
```

### describe the problem

When running `flake8 --help`, the `--count` options is described as:
```
  --count               Print total number of errors and warnings to standard error and set the exit code to 1 if total is not empty.
```

The count is printed to `stdout`, not `stderr`, according to the [documentation](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-count), this [issue](https://github.com/PyCQA/flake8/issues/455), and my own experience.

tmarice:
Fixes #1554


docs/source/manpage.rst
```diff
@@ -38,7 +38,7 @@ All options available as of Flake8 3.1.0::
     -q, --quiet           Report only file names, or nothing. This option is
                           repeatable.
     --count               Print total number of errors and warnings to standard
-                          error and set the exit code to 1 if total is not
+                          output and set the exit code to 1 if total is not
```
asottile, line 41: this isn't meant to be kept up to date (see the text above it) so this can be discarded

in reality I should probably just remove this documentation page and suggest packagers to use help2man or whatever

docs/source/user/invocation.rst
```diff
@@ -92,7 +92,7 @@ And you should see something like:
       -q, --quiet           Report only file names, or nothing. This option is
                             repeatable.
       --count               Print total number of errors and warnings to standard
-                            error and set the exit code to 1 if total is not
```
asottile, line 95: same as above

sigmavirus24 (approved):

asottile (approved):

PR #1561, opened 2022-02-22
Add debug printout when reading config file

fwinkl:
Add a debug printout when reading the (extra) config files(s). A similar line was available in version 3.x before the removal of the user-specific config file parsing.

### DIFF
--- a/src/flake8/options/config.py
+++ b/src/flake8/options/config.py
@@ -61,6 +61,7 @@ def load_config(
 
     cfg = configparser.RawConfigParser()
     if config is not None:
+        LOG.debug("Reading config file %s", config)
         if not cfg.read(config, encoding="UTF-8"):
             raise exceptions.ExecutionError(
                 f"The specified config file does not exist: {config}"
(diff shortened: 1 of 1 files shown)


asottile:
sorry, I'm planning on slowly phasing out debug lines -- use the debugger if you need it

fwinkl:
@asottile, I think there is trade-off to be made between debug messages for "debugging" by developers (i.e. if they have a performance impact) and providing additional information to the user. Especially for projects with multiple sub-directories, possibly using different flake8 configurations, having such an "informational" message can be extremely helpful in trying to understand which configuration file is being used.

The same is true by the way for the current DEBUG message on which error codes are selected/ignored. I hope you are not planning to remove those either.

asottile:
debug messages are not user facing -- if you're depending on them you should seek an alternative

fwinkl:
OK, so what alternative do you suggest for user facing additional information? Pretty much every tool has a --verbose option that allows the more advanced user (not developer) to figure out what is going on if there are problems. Are you suggesting I should just change the level from DEBUG to INFO?

asottile:
if anything it would be added to `--bug-report`

PR #1565, opened 2022-03-10
Syntax for multiple pre-commit additional_dependencies

cclauss:


docs/source/user/using-hooks.rst
```diff
@@ -34,7 +34,7 @@ plugins, use the ``additional_dependencies`` setting.
 .. code-block:: yaml
 
         -   id: flake8
-            additional_dependencies: [flake8-docstrings]
+            additional_dependencies: [flake8-bugbear, flake8-docstrings]
```
sigmavirus24, line 37: I feel like YAML's well enough known that this should be obvious, plus pre-commit does a good job doumenting this (which is why we link to its documentation).

asottile: yep, it's very clearly a list

Issue #1568, opened 2022-03-15
Support hyphens in error codes


jarshwah:
### describe the request

I'd like to be able to use hyphens in error codes so that I can have readable error codes that are name-spaced. For example, here's a list of some error codes we're currently using:

COMPANY-UNIT-TESTS-FILEPATH
COMPANY-FUNCTIONAL-TESTS-FILEPATH
COMPANY-PRINT-STATEMENTS
COMPANY-MOCK-ASSERT-METHODS

These codes work fine with `noqa` directives. The only place they don't seem to work is in flake8 configuration files (such as setup.cfg). That is, the following config file crashes flake8:

```
[flake8]
per-file-ignores =
    tests/test_flake8_plugin.py:COMPANY-UNIT-TESTS-FILEPATH
```

With the error:

```
There was a critical error during execution of Flake8:
Expected `per-file-ignores` to be a mapping from file exclude patterns to ignore codes.

Configured `per-file-ignores` setting:

    tests/test_flake8_plugin.py:COMPANY-UNIT-TESTS-FILEPATH
```

I believe it comes down to this line: https://github.com/PyCQA/flake8/blob/94ed800b2a256dd4feb7e3d680b9c273448f6885/src/flake8/utils.py#L59

The fix could be a simple update of the regex. I will put a patch together shortly to demonstrate a potential fix.

jarshwah:
It sounds like this feature request is rejected, so rather than continue the discussion on the PR I thought I'd try to plead my case here. I'd like to understand the reasons for rejecting this, and wanted to dig in a little deeper to the reasons I think it's a good idea. 

I don't think HYPHENATED-CODES should replace SRT123 codes but there's room for both to exist.

(I also understand that there's no obligation for the maintainers to continue this conversation, and if that's the case, so be it, no worries).

Firstly, the rest of flake8 and the ecosystem of plugins already works without incident with HYPHENATED-CODES. There is code out there in the wild producing error codes that don't match the form SRT123. I could only find the config parser with the original constraints (and a [pycodestyle](https://github.com/PyCQA/pycodestyle/blob/6678922f8d4fb6a34fb77f4f5a7992ade45aaf2f/pycodestyle.py#L129) regex that enforces A123 type codes).

Second, ignoring rules using short-codes impairs readability, as you usually have to go search for what a code means or add comments into the config file to understand what they're used for. Eg 
[...]

PR #1569, opened 2022-03-15
Support hyphens within a flake8 custom plugin code #1568

jarshwah:
Prior to this change, a code had to be formatted as a series of characters
followed by a series of numbers.

This change updates the format to require the first character to be a
letter, followed by any number of letters, numbers, or hyphens.

Implements the feature request from https://github.com/PyCQA/flake8/issues/1568

There are no docs updates yet but I'm happy to update them if the feature request is approved.

### DIFF
--- a/src/flake8/utils.py
+++ b/src/flake8/utils.py
@@ -56,7 +56,7 @@ class _Token(NamedTuple):
 _CODE, _FILE, _COLON, _COMMA, _WS = "code", "file", "colon", "comma", "ws"
 _EOF = "eof"
 _FILE_LIST_TOKEN_TYPES = [
-    (re.compile(r"[A-Z]+[0-9]*(?=$|\s|,)"), _CODE),
+    (re.compile(r"[A-Z]+[A-Z0-9-]*(?=$|\s|,)"), _CODE),
     (re.compile(r"[^\s:,]+"), _FILE),
     (re.compile(r"\s*:\s*"), _COLON),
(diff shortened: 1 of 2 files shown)


src/flake8/utils.py
```diff
@@ -56,7 +56,7 @@ class _Token(NamedTuple):
 _CODE, _FILE, _COLON, _COMMA, _WS = "code", "file", "colon", "comma", "ws"
 _EOF = "eof"
 _FILE_LIST_TOKEN_TYPES = [
-    (re.compile(r"[A-Z]+[0-9]*(?=$|\s|,)"), _CODE),
+    (re.compile(r"[A-Z]+[A-Z0-9-]*(?=$|\s|,)"), _CODE),
```
sigmavirus24, line 59: This may work but it's the wrong change to make.

_If_ we were to accept this, which I don't think we will, the regular expression should be significantly different, something more like:

```pcre
[A-Z][A-Z-]+[0-9]*(?=$|\s|,)
```

sigmavirus24 (changes requested):
I don't think we want to accept this, either as it is or in general

asottile:
agreed

src/flake8/utils.py
jarshwah: Sure completely open to adjusting the regex. But I'm unsure of the need to be so restrictive (numbers can only be at the end for example). I've written up a little more in the feature request "issue".

Issue #284, opened 2021-04-03
Overriding built-in default ignores through an extension


asottile:
In GitLab by @snoack on Jan 10, 2019, 14:17

We'd like to centralize our flake8 configuration through an extension, rather than duplicating the error codes to be selected/ignored in each of our project's `tox.ini`. It's already possible to add default ignores and selects through an extension:

```python
    @classmethod
    def add_options(cls, parser):
        parser.extend_default_ignore(['D1'])
        parser.extend_default_select(['E226'])
```

However, if I select an error code that flake8 ignores by default, the built-in default ignore takes precedence. There is also an `remove_default_ignore()` method but it only removes ignores that have been preiosly added by this plugin, it doesn't have any effect on the built-in default ignores either.

asottile:
In GitLab by @snoack on Jan 10, 2019, 15:09

changed title from **Overriding built-in default ignores though an extension** to **Overriding built-in default ignores th{+r+}ough an extension**

asottile:
In GitLab by @snoack on Jan 10, 2019, 15:12

changed the description

asottile:
In GitLab by @sigmavirus24 on Jan 11, 2019, 04:06

To be brief (because I don't have oodles of time right now to respond, frankly) this isn't what these were designed for. We'd probably want an entirely separate set of APIs for a "plugin" that defines a style guide which we don't have today.

asottile:
In GitLab by @warsaw on Mar 11, 2019, 15:52

I have a similar use case.  I want to add a new code in a new version of my plugin, but I would like to suppress it by default.  Only if a user calls flake8 with `--select=U999` (let's say `999` is my new code), would you get this error reported.  I believe this issue covers the fact that this is not currently possible.

asottile:
In GitLab by @warsaw on Mar 11, 2019, 15:59

mentioned in merge request warsaw/flufl.flake8!1

asottile:
In GitLab by @asottile on Mar 11, 2019, 22:30

One way to do that with the current facilities is to separate the "optional" things into a separate "plugin" (can be in the same distribution)

For instance:

```python
[options.entry_points]
flake8.extension =
    U1 = foo.MainPlugin
    U9 = foo.ExperimentalPlugin
```

And then in your plugin set

```python
class ExperimentalPlugin:
    # ...
    off_by_default = True
```

Then users can opt into this set of optional codes with

```bash
--enable-extension=U9
```

[...]

PR #1609, opened 2022-06-24
don't consider default codes as explicitly selected unless listed

asottile:
see https://github.com/PyCQA/flake8/issues/284#issuecomment-1165962426

### DIFF
--- a/src/flake8/style_guide.py
+++ b/src/flake8/style_guide.py
@@ -47,6 +47,15 @@ class Decision(enum.Enum):
     Selected = "selected error"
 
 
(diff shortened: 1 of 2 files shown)


wyuenho:
@asottile Do you mind making a release? It's been a while and this is quite an important fix.

sigmavirus24:
Don't ask for releases, please. It's rude. Releases will happen as we have time

wyuenho:
Sorry, I'm not aware of how much work is involved for making a release. Is there anything I can help with?

PR #1618, opened 2022-07-13
ignore config files in the home directory

### LINKED ISSUE #1617: Flake8 4.0.1 Behavior Not As Expected Regarding User-Level Configuration Files

```console
ss@gp:~$ sudo -H python3 -m pip install --upgrade flake8
Requirement already satisfied: flake8 in /usr/local/lib/python3.8/dist-packages (4.0.1)
Requirement already satisfied: pyflakes<2.5.0,>=2.4.0 in /usr/local/lib/python3.8/dist-packages (from flake8) (2.4.0)
Requirement already satisfied: pycodestyle<2.9.0,>=2.8.0 in /usr/local/lib/python3.8/dist-packages (from flake8) (2.8.0)
Requirement already satisfied: mccabe<0.7.0,>=0.6.0 in /usr/local/lib/python3.8/dist-packages (from flake8) (0.6.1)
```


```json
{
  "dependencies": [],
  "platform": {
    "python_version": "3.8.10",
    "system": "Linux"
  },
  "plugins": [
    {
      "is_local": false,
      "plugin": "mccabe",
      "version": "0.6.1"
    },
    {
      "is_local": false,
      "plugin": "pycodestyle",
      "version": "2.8.0"
    },
    {
      "is_local": false,
      "plugin": "pyflakes",
      "version": "2.4.0"
    }
  ],
  "version": "4.0.1"
}
```


### describe the problem

#### commands ran
```
ss@gp:~$ uname -a
Linux Gal-Pro 5.4.0-7642-generic #46~1598628707~20.04~040157c~dev-Ubuntu SMP Fri Aug 28 17:55:30  x86_64 x86_64 x86_64 GNU/Linux

ss@gp:~$ lsb_release -id
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.4 LTS

ss@gp:~$ flake8 --version
4.0.1 (mccabe: 0.6.1, pycodestyle: 2.8.0, pyflakes: 2.4.0) CPython 3.8.10 on Linux

ss@gp:~$ cat foo.py 
#!/usr/bin/env python3
"""
This line has 80 characters ....................................................
This line has 79 characters ...................................................
"""

ss@gp:~$ ls -l ~/.config/flake8 ~/.flake8
ls: cannot access '/home/ssiano/.flake8': No such file or directory
-rw-rw-r-- 1 ssiano ssiano 30 Jul 13 11:15 /home/ssiano/.config/flake8

ss@gp:~$ cat ~/.config/flake8
[flake8]
max-line-length = 78

ss@gp:~$ flake8 foo.py 
foo.py:3:80: E501 line too long (80 > 79 characters)

ss@gp:~$ mv ~/.config/flake8 ~/.flake8

ss@gp:~$ flake8 foo.py 
foo.py:3:79: E501 line too long (80 > 78 characters)
foo.py:4:79: E501 line too long (79 > 78 characters)

ss@gp:~$ flake8 --isolated foo.py                                                                                                                                                                                                                           
foo.py:3:80: E501 line too long (80 > 79 characters)
```

#### what I expected to happen
I expected Flake8 4.0.1 to ignore user-level config files, as described [in the Flake8 4.0.0 release notes](https://flake8.pycqa.org/en/latest/release-notes/4.0.0.html#new-dependency-information).

Flake8 did ignore `~/.config/flake8`, but it did not ignore `~/.flake8`.

Also, I thought the `--isolated` option would be removed, but it is still present, and it did result in Flake8 ignoring `~/.flake8`.

asottile:
resolves #1617

### DIFF
--- a/src/flake8/options/config.py
+++ b/src/flake8/options/config.py
@@ -14,7 +14,18 @@ from flake8.options.manager import OptionManager
 LOG = logging.getLogger(__name__)
 
 
+def _stat_key(s: str) -> Tuple[int, int]:
+    # same as what's used by samefile / samestat
+    st = os.stat(s)
+    return st.st_ino, st.st_dev
+
+
 def _find_config_file(path: str) -> Optional[str]:
+    # on windows if the homedir isn't detected this returns back `~`
+    home = os.path.expanduser("~")
+    home_stat = _stat_key(home) if home != "~" else None
+
+    dir_stat = _stat_key(path)
     cfg = configparser.RawConfigParser()
     while True:
         for candidate in ("setup.cfg", "tox.ini", ".flake8"):
@@ -29,10 +40,12 @@ def _find_config_file(path: str) -> Optional[str]:
                     return cfg_path
 
         new_path = os.path.dirname(path)
-        if new_path == path:
+        new_dir_stat = _stat_key(new_path)
+        if new_dir_stat == dir_stat or new_dir_stat == home_stat:
             break
         else:
             path = new_path
+            dir_stat = new_dir_stat
 
     # did not find any configuration file
     return None
--- a/tests/unit/test_options_config.py
+++ b/tests/unit/test_options_config.py
@@ -1,4 +1,6 @@
 import configparser
+import os.path
+from unittest import mock
 
 import pytest
 
@@ -64,6 +66,16 @@ def test_find_config_searches_upwards(tmp_path):
     expected.write_text("[flake8]")
 
     assert config._find_config_file(str(subdir)) == str(expected)
+
+
+def test_find_config_ignores_homedir(tmp_path):
+    subdir = tmp_path.joinpath("d")
+    subdir.mkdir()
+
+    tmp_path.joinpath(".flake8").write_text("[flake8]")
+
+    with mock.patch.object(os.path, "expanduser", return_value=str(tmp_path)):
+        assert config._find_config_file(str(subdir)) is None
 
 
 def test_load_config_config_specified_skips_discovery(tmpdir):


sigmavirus24 (approved):

matthijskooijman:
I came across this PR following a trail from https://stackoverflow.com/a/71974264/740048. I just want to comment that this restriction (ignore `~/.flake8` *and* parent directories) seems rather arbitrary. I believe the reasoning is that always loading `~/.config/flake8` and `~/flake8` was removed in 4.0, but (before this PR), `~/.flake8` would only be loaded if that followed from the look-in-parent-directories policy of finding a project config. If someone does not want `~/.flake8` to be loaded, they can just not create it, right? So IMHO there was no real need for this PR (but changing things now is probably going to cause more fuss, of course).

I guess the underlying goal here is to prevent a .flake8 "project config" from being loaded from outside the project root. Stopping at `~/.flake8` handles this partly, but does not really implement what you need (which is tricky, since there is no uniform definition of a project root I guess).

In any case, I would suggest that this restriction is at least explicitly documented. I believe it is not currently documented, though I could not find any documentation about parent directory traversal at all (I looked at https://flake8.pycqa.org/en/latest/user/configuration.html and https://flake8.pycqa.org/en/latest/internal/option_handling.html#configuration-file-management). Based on what I [previously wrote on stackoverflow](https://stackoverflow.com/a/71974264/740048), some docs about parent traversal do (or did) exist, but I haven't linked to it, so I'm unsure where I read that before (apr 2022).

asottile:
follow the paper trail of this issue and you'll find the original one with the rationale

the docs you link say you can configure flake8 "in your project" -- anywhere else is unsupported

I should have locked this one too -- thanks for the reminder!

Issue #1337, opened 2021-05-19
--jobs Doesn't Work on Mac with Python 3.8+


wbond:
*Please describe how you installed Flake8*

```
$ pipenv install
```

*Please provide the exact, unmodified output of `flake8 --bug-report`*

```
{
  "dependencies": [],
  "platform": {
    "python_implementation": "CPython",
    "python_version": "3.8.5",
    "system": "Darwin"
  },
  "plugins": [
    {
      "is_local": false,
      "plugin": "black",
      "version": "0.1.1"
    },
    {
      "is_local": false,
      "plugin": "flake8-bugbear",
      "version": "21.4.3"
    },
    {
      "is_local": false,
      "plugin": "flake8-print",
      "version": "4.0.0"
    },
    {
      "is_local": false,
      "plugin": "flake8_builtins",
      "version": "1.5.2"
    },
    {
      "is_local": false,
      "plugin": "mccabe",
      "version": "0.6.1"
    },
    {
      "is_local": false,
      "plugin": "pycodestyle",
      "version": "2.7.0"
    },
    {
      "is_local": false,
      "plugin": "pyflakes",
      "version": "2.3.1"
    }
  ],
  "version": "3.9.0"
}
```

*Please describe the problem or feature*

The `--jobs` parameter does nothing on macOS with Python 3.8+.

*If this is a bug report, please explain with examples (and example code) what you expected to happen and what actually happened.*

I ran `flake8 --jobs=8` and it took as long as `flake8 --jobs=1`. It appears that Python 3.8+ on Mac uses `spawn` as the default multiprocessing method. This apparently doesn't work with flake8.

I see this as mostly a documentation bug. Currently the docs say `--jobs` is ignored on Windows, but it should mention Mac and 3.8+.

An alternative, and possibly better, fix would be to error if `--jobs` is specified on a platform that can't respect that arg.

asottile:
fwiw I have plans in the future to rework the plugin system to enable multiprocessing on all platforms which have a multiprocessing module.  but it's a lot of work and I haven't had a ton of freetime to hack on it recently

wbond:
To be honest, the documentation change would still require the vast majority of users to dig into the source code to determine why jobs isn’t working.

asottile:
there is a warning at least:

```console
$ ./venv/Scripts/flake8.exe -v t.py |& grep multiprocessing
[...]

PR #1621, opened 2022-07-24
Enforce usage of fork method on macOS

FlorentJeannot:
Hello there! 👋 

I saw this issue https://github.com/PyCQA/flake8/issues/1337 and I would like to suggest this quick fix to restore the multiprocessing module for macOS while we wait for this https://github.com/PyCQA/flake8/issues/1337#issuecomment-1040277568.

I am aware that an issue with macOS and fork has been found by the Python core developers, but I do not recall having any issues using flake8 on my MacBook with Python 3.7.

I am using Python 3.9 on a big project and running flake8 takes about 1m 20s without the multiprocessing module. With this fix, it now takes about 24s.

@asottile Do you think it would be reasonable to allow macOS users to run flake8 with the fork method while we wait for the new implementation? If we shouldn't restore it by default for all users, could we add a temporary option to the command line (something like `--macos-enable-jobs`)?

Thanks!

### DIFF
--- a/src/flake8/checker.py
+++ b/src/flake8/checker.py
@@ -5,6 +5,7 @@ import errno
 import itertools
 import logging
 import multiprocessing.pool
+import platform
 import signal
 import tokenize
 from typing import Any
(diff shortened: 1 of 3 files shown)


asottile:
nah fork is pretty broken on macos

PR #1628, opened 2022-07-31
Remove needless sort in `_style_guide_for`

mxr:
We are always returning the last element so a 'max' operation is sufficient instead of sorting. Note the old code did not handle an empty list so this change doesn't either


asottile (approved):

mxr:
🎵 I miss the old Tony, posts cat gifs when you merge Tony 🎶

asottile:
> 🎵 I miss the old Tony, posts cat gifs when you merge Tony 🎶

my script is broken :'(  -- stupid SPAs

PR #1633, opened 2022-07-31
require python>=3.7

asottile:


asottile:
will merge in a weekish after I'm not going to make any more 5.0.x releases

PR #1634, opened 2022-08-01
Extracted list of config filename so that it can be overriden

wieczorek1990:
Please see related:
https://github.com/csachs/pyproject-flake8/pull/14

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 .coverage
 .coverage.*
 .eggs
+.idea
 .tox
 /.mypy_cache
 build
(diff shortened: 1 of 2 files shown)


sadikkuzu:
Hi @wieczorek1990 

I may miss something.
What is the difference between:

```python
for candidate in ("setup.cfg", "tox.ini", ".flake8"):
```

and

```python
CONFIG_FILENAMES = ("setup.cfg", "tox.ini", ".flake8")
for candidate in CONFIG_FILENAMES:
```

?

Thanks, have a nice day.

wieczorek1990:
This way some other package can extend the list. No easy way to extend a tuple used in a function.
Same usage before version 5:
https://github.com/csachs/pyproject-flake8/blob/main/pflake8/__init__.py#L57

asottile:
monkeypatching is not supported

PR #1646, opened 2022-08-01
Clarify entry point naming

televi:
Clarified what is and is not a valid entry point name
for registering plugins.

### DIFF
--- a/docs/source/plugin-development/registering-plugins.rst
+++ b/docs/source/plugin-development/registering-plugins.rst
@@ -112,10 +112,16 @@ look like::
 
     X101 = flake8_example:ExamplePlugin
 
+In the above case, the entry-point name and the error code produced by your
+plugin are the same.
+
 If your plugin reports several error codes that all start with ``X10``, then
 it would look like::
 
     X10 = flake8_example:ExamplePlugin
+
+In this casae as well as the following case, your entry-point name acts as
+a prefix to the error codes produced by your plugin.
 
 If all of your plugin's error codes start with ``X1`` then it would look
 like::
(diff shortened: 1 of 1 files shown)


docs/source/plugin-development/registering-plugins.rst
```diff
@@ -112,6 +112,9 @@ look like::
 
     X101 = flake8_example:ExamplePlugin
 
+In the above, ``X`` is the entry point prefix and ``X101`` is the
+entry point name.
```
asottile, line 116: ah so actually both `X` and `X101` are used as the entry point name (which *is* the prefix) -- the error code is `X101`

televi: Looks like I wasn't fully understanding that still.  I was looking at "ABC" and "123" as 2 different parts of the entry point name.  I saw how the combination was the error code (or error code "prefix" in the case of something like X1 or just X).  I added this to try to capture your comment about 3 letters being the max prefix length - I was trying to define what the "prefix" meant for an entry point.  Would "In the above, you can think of the letters as a prefix to the entry point name." be better or maybe just leave those two lines out?

asottile: the entry point is the prefix of the codes that the plugin produces -- I think that's the important part from this section

televi: Ah - that's a much cleaner way to express it!  I'll push an update that (hopefully) reflects that cleanly.

asottile (approved):

televi:
Thank you so much for your help and great comments/insights!  I love the cat gif!

asottile:
> Thank you so much for your help and great comments/insights! I love the cat gif!

heh -- I have a user script which auto-posts them https://www.youtube.com/watch?v=2rDNJvHznrM

Issue #1661, opened 2022-08-10
flake8 color config not picked up in either .flake8, setup.cfg, or tox.ini


velicanu:
### how did you install flake8?

```console
pip install flake8  # currently version 5.0.4
```

### unmodified output of `flake8 --bug-report`

```json
{
  "platform": {
    "python_implementation": "CPython",
    "python_version": "3.9.13",
    "system": "Darwin"
  },
  "plugins": [
    {
      "plugin": "mccabe",
      "version": "0.7.0"
    },
    {
      "plugin": "pycodestyle",
      "version": "2.9.1"
    },
    {
      "plugin": "pyflakes",
      "version": "2.5.0"
    }
  ],
  "version": "5.0.4"
}
```

#### what I expected to happen
Added a `color = never` to the .flake8 config in my project and expected the color output to be disabled. I confirmed that the config file itself was being picked up by toggling the warnings flake8 produced. I also tried with both `setup.cfg` and `tox.ini` but the color config isn't picked up in those either.

I stumbled across this because I suspect that color defaulting to `on` in 5.0 is responsible for breaking [flycheck](https://github.com/flycheck/flycheck/issues/1969) compatibility.

#### sample code

Full .flake8 config
```
[flake8]
ignore = E203, E266, E501, W503
max-complexity = 18
color = never
```

Python code
```python
def bar():
    x = "123"

    json.dumps({})
```

#### flake8 output

<img width="598" alt="Screen Shot 2022-08-10 at 8 03 01 AM" src="https://user-images.githubusercontent.com/5022772/183938599-d2ba9fa6-ad79-4455-bc30-391f706fd5c3.png">

asottile-sentry:
I don't think this is what's breaking flycheck -- it looks like they execute always with `--apend-config .flake8rc` which will now error if `.flake8rc` does not exist

they should adjust their call to not pass configuration files which don't exist

asottile:
I don't think this is what's breaking flycheck -- it looks like they execute always with `--apend-config .flake8rc` which will now error if .flake8rc does not exist

they should adjust their call to not pass configuration files which don't exist

PR #1662, opened 2022-08-10
fix documentation of color configuruation

velicanu:
This PR fixes https://github.com/PyCQA/flake8/issues/1661 , where the `color` option was not being picked up from the config file, which I assumed was a bug since the documentation suggests it should be configurable from there: https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-color

### DIFF
--- a/docs/source/user/options.rst
+++ b/docs/source/user/options.rst
@@ -193,7 +193,7 @@ Options and their Descriptions
 
     Possible options are ``auto``, ``always``, and ``never``.
 
-    This **can** be specified in config files.
+    This **can not** be specified in config files.
 
     When color is enabled, the following substitutions are enabled:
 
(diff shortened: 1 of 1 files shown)


velicanu:
Closes #1661

asottile:
hmmm actually the docs are probably wrong here and this shouldn't be configurable

asottile-sentry:
can you fix the docs instead? thanks!

asottile:
can you fix the docs instead? thanks!

asottile:
in the future please don't make PRs from your `main` branch as it makes it significantly more difficult to review / fix

Cheaterman:
Is there any reason to prevent this from being set in config?

PR #1667, opened 2022-08-14
Ensures help text defaults are not clobbered by config values

### LINKED ISSUE #1666: Help text claims custom config values to be the default values

```console
pyenv install 3.9.13
pyenv local 3.9.13
pyenv virtualenv flake8-py39
pyenv activate flake8-py39
pip install -U pip
pip install flake8
```

```json
{
  "platform": {
    "python_version": "3.9.13",
    "system": "Darwin"
  },
  "plugins": [
    {
      "plugin": "mccabe",
      "version": "0.7.0"
    },
    {
      "plugin": "pycodestyle",
      "version": "2.9.1"
    },
    {
      "plugin": "pyflakes",
      "version": "2.5.0"
    }
  ],
  "version": "5.0.4"
}
```

### What I expected to happen
<br/>

Let's say I have a `.flake8` config file such as:

```
[flake8]
exclude = 1,2
```

When I run `flake8 -h`, I would expect the help text for the `--exclude` option to say the following (note the defaults):

```
  --exclude patterns    Comma-separated list of files or directories to exclude. (Default: ['.svn', 'CVS', '.bzr', '.hg', '.git', '__pycache__', '.tox', '.nox', '.eggs', '*.egg'])
```

However, the actual output clobbers the true defaults and instead claims the config values as the default values (which is quite unintuitive and can be very misleading if you take them to be the actual default values that flake8 provides):

```
  --exclude patterns    Comma-separated list of files or directories to exclude. (Default: ['1', '2'])
```

One other trivial thing is that the current output of comma-separated lists does not match the format shown in the [documentation](https://flake8.pycqa.org/en/latest/user/invocation.html#invoking-flake8): 

`['.svn', 'CVS', '.bzr', '.hg', '.git', '__pycache__', '.tox', '.nox', '.eggs', '*.egg']` 
vs. the documented:
`.svn,CVS,.bzr,.hg,.git,__pycache__,.tox,.nox,.eggs,.egg`.

  
### Why it happens
<br/>

Looking at `src/flake8/options/aggregator.py`, the `aggregate_options` function overrides the default option values with the config values. Moreover, in `src/flake8/main/options.py`, we're using the `%(default)s` format specifier in the option help strings. Thus, when the help text actually gets printed it uses the updated values instead of the original defaults 😢 .

  
### Failing test case
<br/>

Adding this test case to `tests/integration/test_main.py` asserts the desired behavior and successfully reproduces the issue:

```python
import re

from flake8.defaults import EXCLUDE


def test_config_value_does_not_clobber_default_help_text(tmpdir, capsys):
    """Test a config value does not clobber the default help text output."""
    setup_cfg = """\
[flake8]
exclude = 1,2
"""
    expected = ",".join(EXCLUDE)

    default_pattern = re.compile(r"\(Default:(.*?)\)")
    whitespace_pattern = re.compile(r"\s+")

    with tmpdir.as_cwd():
        tmpdir.join("setup.cfg").write(setup_cfg)
        with pytest.raises(SystemExit):
            cli.main(["-h"])

    out, err = capsys.readouterr()

    for section in out.split("--"):
        if section.startswith("exclude patterns"):
            section = re.sub(whitespace_pattern, "", section)
            default_groups = re.search(default_pattern, section)
            assert default_groups is not None
            default = default_groups.group(1)
            assert default == expected
```

  
### Potential solution
<br/>

I believe the simplest solution would be to handle this in the initializer for the `Option` class (`src/flake8/options/manager.py`), binding any instance of `%default` or `%(default)s` in the `help` arg with the passed `default` arg, such that it doesn't subsequently get overridden by later argparse shenanigans. This should also fix the difference in format vs. the documentation that I mentioned.

If there's an appetite for it, **I would love to put together a PR** with the above test/implementation. Thanks!

chrisRedwine:
This resolves #1666

### DIFF
--- a/src/flake8/options/manager.py
+++ b/src/flake8/options/manager.py
@@ -204,7 +204,9 @@ class Option:
             help = help.replace(_optparse_default, _argparse_default)
 
         # ensure help text doesn't get clobbered by later updates to defaults
-        help = self._bind_defaults_to_help_text(help, default)
+        help = self._bind_defaults_to_help_text(
+            help, default, long_option_name
+        )
 
         # optparse -> argparse for `callback`
         if action == "callback":
--- a/src/flake8/options/manager.py
+++ b/src/flake8/options/manager.py
@@ -36,6 +36,10 @@ _optparse_callable_map: Dict[str, Union[Type[Any], _ARG]] = {
 }
 
 
+_optparse_default: str = "%default"
+_argparse_default: str = "%(default)s"
+
+
 class _CallbackAction(argparse.Action):
     """Shim for optparse-style callback actions."""
 
@@ -189,13 +193,18 @@ class Option:
             short_option_name, long_option_name = _ARG.NO, short_option_name
 
         # optparse -> argparse `%default` => `%(default)s`
-        if help is not _ARG.NO and "%default" in help:
+        if help is not _ARG.NO and _optparse_default in help:
             LOG.warning(
-                "option %s: please update `help=` text to use %%(default)s "
-                "instead of %%default -- this will be an error in the future",
+                "option %s: please update `help=` text to use %s "
+                "instead of %s -- this will be an error in the future",
                 long_option_name,
+                _argparse_default,
+                _optparse_default,
             )
-            help = help.replace("%default", "%(default)s")
+            help = help.replace(_optparse_default, _argparse_default)
+
+        # ensure help text doesn't get clobbered by later updates to defaults
+        help = self._bind_defaults_to_help_text(help, default)
 
         # optparse -> argparse for `callback`
         if action == "callback":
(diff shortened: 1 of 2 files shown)


chrisRedwine:
Tested via:
`tox -e py`
`tox -e pre-commit`
`tox -e linters`

You can also see that the help text output:
Screen Shot 2022-08-14 at 2 48 49 AM

Now matches the [docs](https://flake8.pycqa.org/en/latest/user/invocation.html):
Screen Shot 2022-08-14 at 2 53 04 AM

chrisRedwine:
I just noticed there are some other minor differences in the format that the docs show for `--help` output vs. actual. I'm happy to sync the docs with the current format, if that's desired (or create a different issue for that) - just let me know.

asottile:
the help containing your values is intentional -- the configuration is the user overriding the defaults

the docs you're referring to are intentionally not kept up to date ("something like") and the optparse shim will be removed (it is a temporary bridge) so we should not give it more features

chrisRedwine:
Thanks for the quick response!

Ok cool, having the docs vague on the output so there's not a need to constantly keep them in sync makes total sense.

We were really surprised to hear that the help text behavior was intentional, though. That doesn't seem to match what the definition of a default is. E.g., if I pass an arg `42` to a function where the default is `0`, the default value of that arg is _still_ `0` regardless of what I pass in, and I'd expect the docs to reflect that, I think. If I could (politely) push my luck, I'd ask to just reconsider the output there by way of an example that actually bit us:

We were reviewing the flake8 configs across our repos, and were wanting to add a value to our `exclude` list, which was `.git,venv,.venv,setup.py,*.sql,*.lock` or something. Running `help`, we saw the text say `(Default: ...)` with our exact values and thought "Oh awesome, flake8 tells us what its default configuration is for its options! Looks like we can just remove our `exclude` and take advantage of `extend-exclude` to add our single new value without having to overwrite all the defaults!" Of course, then we got errors in CI that took a while to track down and realize the defaults weren't what we thought they were.

Anyways, I really appreciate your work on these great projects - been enjoying the streams, btw!

the-wondersmith:
Just chiming in, I'd have to second @chrisRedwine's request for the _default_ values to always be displayed as the _actual_ default values. I've personally run into the confusion he referenced just by running `flake8 --help` in two different projects on my local drive.

It's a little surreal to do:
```bash
cd ~/Projects/some-project
flake8 --help

cd ~/Projects/some-other-project
flake8 --help
```
and see two _obviously_ different outputs. Literally made me chase down python and flake 8 versions for nearly half an hour 🥲

asottile:
you can use `--isolated`

chrisRedwine:
Well, I think that's extremely non-obvious, unintuitive behavior for a CLI, but I do genuinely appreciate the time. Thanks again.

PR #1669, opened 2022-08-18
include bin/ subdirectory in sdist tarballs

dirkmueller:
without that the testsuite is failing on the missing files


asottile:
we don't intend to test from the sdist -- use the github source tgz

PR #1674, opened 2022-09-03
check `isclass` instead of `not isfunction` when gathering plugin parameters

rmorshea:
It is not always true that `not isfunction(...) is isclass(...)`.

### DIFF
--- a/src/flake8/plugins/finder.py
+++ b/src/flake8/plugins/finder.py
@@ -272,7 +272,7 @@ def _parameters_for(func: Any) -> dict[str, bool]:
         A dictionary mapping the parameter name to whether or not it is
         required (a.k.a., is positional only/does not have a default).
     """
-    is_class = not inspect.isfunction(func)
+    is_class = inspect.isclass(func)
     if is_class:
         func = func.__init__
 


asottile:
flake8 only supports plugins which are functions or classes

do you have an actual bug you're trying to fix?

rmorshea:
I suppose it depends if you'd consider objects with `__call__` methods to be "functions". My intuition is that this would be allowed since actual functions and objects with `__call__` methods behave identically. From what I can tell, `flake8` gracefully handles such objects under this change.

rmorshea:
In that case it seems like `flake8` should raise an error if `not (isclass(...) or isfunction(...))`. Right now, `flake8` behaves strangely and gives confusing error messages.

asottile:
unless you have an actual bug and/or a test this is not a helpful patch and looks like you're just trying to boost your commit stats.  you also didn't answer my question which further indicates to me you're not actually trying to fix something.  instances with `__call__` also won't work because their argspec contains `self`

rmorshea:
The documentation states:

> Flake8 now does not distinguish between the two types of plugins. Any plugin can accept either an AST or a range of arguments. Further, any plugin that has certain callable attributes can also register options and receive parsed options.

The docs further explain that:

> Any plugin that has callable attributes add_options and parse_options can parse option information and register new options.

I interpreted this to mean that I could write a plugin like:

```python
class MyPlugin:
    def add_options(self, option_manager): ...
    def parse_option(self, options): ...
    def __call__(self, tree): ...

plugin = MyPlugin()
```

Also, `inspect.signature` correctly handles methods like `__call__` and excludes `self` from the parameter list:

```python
import inspect

class X:
    def __call__(self, a, b, c): ...

assert list(inspect.signature(X()).parameters) == ["a", "b", "c"]
```

My reason for wanting to define a plugin like this is that make testing my plugin easier. Consider the following class-based plugin:

```python
class MyPlugin:
    @classmethod
    def parse_option(cls, options):
        if options.some_option:
            cls.an_attribute = ...
        else:
            cls.a_different_attribute = ...
```

If in one test cases `options.some_option is True` and in the next `options.some_option is False` I'll have a problem where state persists across test cases because, after the first test, `MyPlugin.an_attribute` will still exist. Were I able to define my plugin using an instance I could isolate state much more easily by simply creating a new plugin instance for each test. Perhaps there's a way to avoid this problem in a way I haven't thought of though.

Ultimately though, what lead me to write this patch is the fact that I was confused by the error message `flake8` gave me since it was complaining that it didn't know what parameters to pass my plugin. When I traced the problem to `is_class = not isfunction(...)` it seemed like `isclass` matched up better with what I believed the intended behavior was.

I should have laid this all out in the description, but I also didn't think this would be a controversial change.

asottile:
please understand a PR made from the github UI which changes behaviour with no test and no rationale looks like spam -- this is a popular repository so we get a lot of these sort of "improves nothing" drive-by patches.  in addition not answering my direct question makes it look even more like you're churning just to churn.  even if it wasn't a popular repository you should always be explaining the why of your change, preferably with an automated test demonstrating your change -- this isn't specific to flake8 this is just being a good open source citizen.

I'm still not going to change this as I'm planning a revamp of plugins to not be global variables and expanding the surface area by extending them to another type of object will only make that migration more difficult

rmorshea:
Apologies, I definitely could have gone about this better (e.g. filing this as an issue first).

Issue #1704, opened 2022-10-02
Configuration files are silently ignored


dHannasch:
### how did you install flake8?

```console
pip install flake8
Collecting flake8
  Downloading flake8-5.0.4-py2.py3-none-any.whl (61 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 61.9/61.9 kB 9.4 MB/s eta 0:00:00
Collecting mccabe<0.8.0,>=0.7.0
  Downloading mccabe-0.7.0-py2.py3-none-any.whl (7.3 kB)
Collecting pyflakes<2.6.0,>=2.5.0
  Downloading pyflakes-2.5.0-py2.py3-none-any.whl (66 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 66.1/66.1 kB 21.3 MB/s eta 0:00:00
Collecting pycodestyle<2.10.0,>=2.9.0
  Downloading pycodestyle-2.9.1-py2.py3-none-any.whl (41 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 41.5/41.5 kB 11.7 MB/s eta 0:00:00
Installing collected packages: pyflakes, pycodestyle, mccabe, flake8
Successfully installed flake8-5.0.4 mccabe-0.7.0 pycodestyle-2.9.1 pyflakes-2.5.0
```

### unmodified output of `flake8 --bug-report`

```json
{
  "platform": {
    "python_implementation": "CPython",
    "python_version": "3.10.7",
    "system": "Linux"
  },
  "plugins": [
    {
      "plugin": "mccabe",
      "version": "0.7.0"
    },
    {
      "plugin": "pycodestyle",
      "version": "2.9.1"
    },
    {
      "plugin": "pyflakes",
      "version": "2.5.0"
    }
  ],
  "version": "5.0.4"
}
```

### describe the problem

Configuration files can be silently ignored: https://github.com/dHannasch/test-flake8

[...]

PR #1705, opened 2022-10-02
Emit error on invalid config

dHannasch:
Emit an error message if the configuration is invalid due to multiple
configuration files.

This shows how we can easily emit an error in the simplest case,
though we can also use _walk_up_directories_to_root_or_home to make this more
robust.

Resolves https://github.com/PyCQA/flake8/issues/1704.


asottile:
before working on a feature please confirm it is desirable first

this is not as it is slower to discover configuration and is also breaking

PR #1713, opened 2022-10-12
Adds error when incorrect error codes are parsed from config file

### LINKED ISSUE #1689: warn on invalid codes in `ignore` / `extend-ignore`
### describe the request

for example https://github.com/getsentry/sentry-python/blob/b36d84a76bd6f8344c9b0a9694591939296e9c06/.flake8#L2-L13

this disables a code called `//` and a code called `I` (among many others) due to incorrect comment syntax

mennoliefstingh:
This PR resolves #1689 by raising a warning when parsing a configuration file results in values for the --ignore or --extend-ignore options that cannot be valid error codes, i.e. they are not in a 'A123' format.

### DIFF
--- a/src/flake8/options/config.py
+++ b/src/flake8/options/config.py
@@ -7,6 +7,7 @@ import os.path
 from typing import Any
 
 from flake8 import exceptions
+from flake8.defaults import VALID_CODE_PREFIX
 from flake8.options.manager import OptionManager
 
 LOG = logging.getLogger(__name__)
@@ -120,6 +121,16 @@ def parse_config(
         LOG.debug('Option "%s" returned value: %r', option_name, value)
 
         final_value = option.normalize(value, cfg_dir)
+
+        if option_name in {"ignore", "extend-ignore"}:
+            for error_code in final_value:
+                if not VALID_CODE_PREFIX.match(error_code):
+                    raise ValueError(
+                        f"Error code {error_code!r} "
+                        f"supplied to {option_name!r} option "
+                        f"does not match {VALID_CODE_PREFIX.pattern!r}"
+                    )
+
         assert option.config_name is not None
         config_dict[option.config_name] = final_value
 
--- a/tests/unit/plugins/finder_test.py
+++ b/tests/unit/plugins/finder_test.py
@@ -31,37 +31,6 @@ def _loaded(plugin=None, obj=None, parameters=None):
     return finder.LoadedPlugin(plugin, obj, parameters)
 
 
-@pytest.mark.parametrize(
-    "s",
-    (
-        "E",
-        "E1",
-        "E123",
-        "ABC",
-        "ABC1",
-        "ABC123",
-    ),
-)
-def test_valid_plugin_prefixes(s):
-    assert finder.VALID_CODE.match(s)
-
-
-@pytest.mark.parametrize(
-    "s",
-    (
-        "",
-        "A1234",
-        "ABCD",
-        "abc",
-        "a-b",
-        "☃",
-        "A𝟗",
-    ),
-)
-def test_invalid_plugin_prefixes(s):
-    assert finder.VALID_CODE.match(s) is None
-
-
 def test_loaded_plugin_entry_name_vs_display_name():
     loaded = _loaded(_plugin(package="package-name", ep=_ep(name="Q")))
     assert loaded.entry_name == "Q"
--- a/tests/unit/test_options_config.py
+++ b/tests/unit/test_options_config.py
@@ -220,3 +220,35 @@ def test_parse_config_ignores_unknowns(tmp_path, opt_manager, caplog):
 def test_load_config_missing_file_raises_exception(capsys):
     with pytest.raises(exceptions.ExecutionError):
         config.load_config("foo.cfg", [])
+
+
+def test_invalid_ignore_codes_raise_error(tmpdir, opt_manager):
+    tmpdir.join("setup.cfg").write("[flake8]\nignore = E203, //comment")
+    with tmpdir.as_cwd():
+        cfg, _ = config.load_config("setup.cfg", [], isolated=False)
+
+    with pytest.raises(ValueError) as excinfo:
+        config.parse_config(opt_manager, cfg, tmpdir)
+
+    expected = (
+        "Error code '//comment' supplied to 'ignore' option "
+        "does not match '^[A-Z]{1,3}[0-9]{0,3}$'"
+    )
+    (msg,) = excinfo.value.args
+    assert msg == expected
+
+
+def test_invalid_extend_ignore_codes_raise_error(tmpdir, opt_manager):
+    tmpdir.join("setup.cfg").write("[flake8]\nextend-ignore = E203, //comment")
+    with tmpdir.as_cwd():
+        cfg, _ = config.load_config("setup.cfg", [], isolated=False)
+
+    with pytest.raises(ValueError) as excinfo:
+        config.parse_config(opt_manager, cfg, tmpdir)
+
+    expected = (
+        "Error code '//comment' supplied to 'extend-ignore' option "
+        "does not match '^[A-Z]{1,3}[0-9]{0,3}$'"
+    )
+    (msg,) = excinfo.value.args
+    assert msg == expected
--- a/src/flake8/defaults.py
+++ b/src/flake8/defaults.py
@@ -43,3 +43,5 @@ NOQA_INLINE_REGEXP = re.compile(
 )
 
 NOQA_FILE = re.compile(r"\s*# flake8[:=]\s*noqa", re.I)
+
+VALID_CODE_PREFIX = re.compile("^[A-Z]{1,3}[0-9]{0,3}$", re.ASCII)
--- a/src/flake8/plugins/finder.py
+++ b/src/flake8/plugins/finder.py
@@ -5,7 +5,6 @@ import configparser
 import inspect
 import itertools
 import logging
-import re
 import sys
 from typing import Any
 from typing import Generator
@@ -14,12 +13,11 @@ from typing import NamedTuple
 
 from flake8 import utils
 from flake8._compat import importlib_metadata
+from flake8.defaults import VALID_CODE_PREFIX
 from flake8.exceptions import ExecutionError
 from flake8.exceptions import FailedToLoadPlugin
 
 LOG = logging.getLogger(__name__)
-
-VALID_CODE = re.compile("^[A-Z]{1,3}[0-9]{0,3}$", re.ASCII)
 
 FLAKE8_GROUPS = frozenset(("flake8.extension", "flake8.report"))
 
@@ -337,10 +335,10 @@ def _classify_plugins(
             raise NotImplementedError(f"what plugin type? {loaded}")
 
     for loaded in itertools.chain(tree, logical_line, physical_line):
-        if not VALID_CODE.match(loaded.entry_name):
+        if not VALID_CODE_PREFIX.match(loaded.entry_name):
             raise ExecutionError(
                 f"plugin code for `{loaded.display_name}` does not match "
-                f"{VALID_CODE.pattern}"
+                f"{VALID_CODE_PREFIX.pattern}"
             )
 
     return Plugins(
--- a/tests/unit/test_defaults.py
+++ b/tests/unit/test_defaults.py
@@ -1,0 +1,36 @@
+from __future__ import annotations
+
+import pytest
+
+from flake8.defaults import VALID_CODE_PREFIX
+
+
+@pytest.mark.parametrize(
+    "s",
+    (
+        "E",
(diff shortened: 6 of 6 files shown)


src/flake8/options/aggregator.py
```diff
@@ -8,9 +8,11 @@
 import argparse
 import configparser
 import logging
+import re
 from typing import Sequence
 
 from flake8.options import config
+from flake8.options.config import _find_config_file
```
asottile, line 15: private names should not be imported (also it's possible the values below don't come from those config files but instead from plugins -- so I don't think it's right to show the filename anyway)

src/flake8/options/aggregator.py
```diff
@@ -43,6 +45,16 @@ def aggregate_options(
             assert isinstance(dest_val, str)
             dest_name = dest_val
 
+        if config_name in ["ignore", "extend-ignore"]:
+            for error_code in value:
+                if not re.match("^[A-Z][0-9]{3}$", error_code):
```
asottile, line 50: this should reuse the same pattern that's used for validating entrypoints, notably they can be any error prefix here

src/flake8/options/aggregator.py
```diff
@@ -43,6 +45,16 @@ def aggregate_options(
             assert isinstance(dest_val, str)
             dest_name = dest_val
 
+        if config_name in ["ignore", "extend-ignore"]:
```
asottile, line 48: should use a set containment rather than list containment here -- also I think this might be too late in the process -- would prefer this get validated during argument / config parsing if possible

asottile (commented):
will need a test as well

I think this should probably _error_ rather than warn since the configuration is just incorrect

src/flake8/options/aggregator.py
mennoliefstingh: good point, listing the invalid error code in the warning should already give the user enough information to know where it's coming from

src/flake8/options/aggregator.py
mennoliefstingh: Also a good point, there is an earlier opportunity to check this immediately after option values have been normalized to a list of strings in config.py

src/flake8/options/aggregator.py
asottile: I don't like repeating myself but that's not right -- it should reuse the exact same pattern that's used for validating entrypoints

there should be no commas or spaces when it checks it

asottile:
> Would you say instead of logging a warning we'd want to raise an exception then? E.g. like passing a non-existent config file raises an ExecutionError:
> 
> ```
> cfg = configparser.RawConfigParser()
>     if config is not None:
>         if not cfg.read(config, encoding="UTF-8"):
>             raise exceptions.ExecutionError(
>                 f"The specified config file does not exist: {config}"
>             )
> ```

ideally it would be early enough in the validation that it would raise the argument parsing error rather than `ExecutionError`

src/flake8/options/aggregator.py
mennoliefstingh: Ooh wait you're referring to `VALID_CODE = re.compile("^[A-Z]{1,3}[0-9]{0,3}$", re.ASCII)`. 
I missed that initially, sorry.

mennoliefstingh:
All right, fixes in these last few commits: 

- Updated the regex pattern to the one used for entry point validation
- Moved argument validation further up in the process, i.e. immediately after the option values are parsed to a list
- Invalid codes now throw a ParsingError instead of logging a warning
- Wrote a unit test that tests with example invalid input (as in issue #1689) for both ignore and extend-ignore

mennoliefstingh:
whoops except for one small mistake, let me fix it right quick

src/flake8/options/config.py
```diff
@@ -120,7 +124,20 @@ def parse_config(
...
+                        f"does not match {VALID_CODE.pattern}"
+                    )
+
         assert option.config_name is not None
         config_dict[option.config_name] = final_value
 
     return config_dict
+
+
+VALID_CODE = re.compile("^[A-Z]{1,3}[0-9]{0,3}$", re.ASCII)
```
asottile, line 143: you've got this defined in a couple places (3?) -- probably want to bring them all together

mennoliefstingh: Good idea! I imported the constant from finder.py now -- I considered moving it upwards into utils.py but decided against it with the contribution guidelines in mind (no extraneous refactoring), but you could make a case that it's better to import downwards (both import from utils.py) than sideways as we do now. What are your thoughts on this?

src/flake8/options/config.py
```diff
@@ -120,6 +121,16 @@ def parse_config(
         LOG.debug('Option "%s" returned value: %r', option_name, value)
 
         final_value = option.normalize(value, cfg_dir)
+
+        if option_name in {"ignore", "extend-ignore"}:
+            for error_code in final_value:
+                if not VALID_CODE.match(error_code):
+                    raise configparser.ParsingError(
```
asottile, line 128: I don't think this is the right type -- the others raise `ValueError` so we should probably do similarly?

src/flake8/options/config.py
```diff
@@ -96,7 +97,7 @@ def parse_config(
     cfg: configparser.RawConfigParser,
     cfg_dir: str,
 ) -> dict[str, Any]:
-    """Parse and normalize the typed configuration options."""
+    """Parse and normalize the typed configuration options"""
```
asottile, line 100: not sure why this changed?

src/flake8/options/config.py
```diff
@@ -8,6 +8,7 @@
 
 from flake8 import exceptions
 from flake8.options.manager import OptionManager
+from flake8.plugins.finder import VALID_CODE
```
asottile, line 11: let's put this into flake8.defaults and call it `VALID_CODE_PREFIX`

tests/unit/test_options_config.py
```diff
@@ -218,3 +218,15 @@ def test_parse_config_ignores_unknowns(tmp_path, opt_manager, caplog):
...
+@pytest.mark.parametrize("option", ("ignore", "extend-ignore"))
+def test_invalid_ignore_codes_raise_error(tmpdir, opt_manager, option):
+    tmpdir.join("setup.cfg").write(
+        f"[flake8]\n{option} = E203, // this is invalid comment syntax"
+    )
+    with tmpdir.as_cwd():
+        cfg, _ = config.load_config("setup.cfg", [], isolated=False)
+
+    with pytest.raises(configparser.ParsingError):
+        _ = config.parse_config(opt_manager, cfg, tmpdir)
```
asottile, line 232: this assignment does nothing

tests/unit/test_options_config.py
```diff
@@ -218,3 +218,15 @@ def test_parse_config_ignores_unknowns(tmp_path, opt_manager, caplog):
...
+
+@pytest.mark.parametrize("option", ("ignore", "extend-ignore"))
+def test_invalid_ignore_codes_raise_error(tmpdir, opt_manager, option):
+    tmpdir.join("setup.cfg").write(
+        f"[flake8]\n{option} = E203, // this is invalid comment syntax"
+    )
+    with tmpdir.as_cwd():
+        cfg, _ = config.load_config("setup.cfg", [], isolated=False)
+
+    with pytest.raises(configparser.ParsingError):
```
asottile, line 231: if you're raising a specific exception with a specific message that should be tested:

```python
with pytest.raises(...) as excinfo:
    ...
msg, = excinfo.value.args
assert msg == ...
```

src/flake8/options/config.py
mennoliefstingh: e.g. supplying an undecodable config file throws a ParsingError (```Raised when a configuration file does not follow legal syntax.```), but I agree that in this context it makes more sense to raise a ValueError since it'll also be raised if a user supplies an invalid code in otherwise valid syntax.

tests/unit/test_options_config.py
```diff
@@ -218,3 +219,25 @@ def test_parse_config_ignores_unknowns(tmp_path, opt_manager, caplog):
 def test_load_config_missing_file_raises_exception(capsys):
     with pytest.raises(exceptions.ExecutionError):
         config.load_config("foo.cfg", [])
+
+
+@pytest.mark.parametrize("option", ("ignore", "extend-ignore"))
+def test_invalid_ignore_codes_raise_error(tmpdir, opt_manager, option):
+    invalid_option = "//comment"
```
asottile, line 226: DRY here is bad -- it's much harder to understand what's going on in the test -- https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

tests/unit/plugins/finder_test.py
```diff
@@ -43,7 +44,7 @@ def _loaded(plugin=None, obj=None, parameters=None):
     ),
 )
 def test_valid_plugin_prefixes(s):
-    assert finder.VALID_CODE.match(s)
+    assert VALID_CODE_PREFIX.match(s)
```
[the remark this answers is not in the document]
mennoliefstingh: Good point, I moved them to a new `test_error_code_validation()` module.

tests/unit/test_options_config.py
mennoliefstingh: Cool, never thought about the fact that KISS is more important than DRY for testing. Learning a lot here 😄 really appreciate the comments.

I removed the `invalid_option` variable but did keep the others as they're imported from elsewhere or originate from the `mark.parametrize` decorator.

tests/unit/test_error_code_validation.py
```diff
@@ -0,0 +1,36 @@
+from __future__ import annotations
```
asottile, line 1: this module should be named after the module it tests

tests/unit/test_options_config.py
```diff
@@ -218,3 +219,21 @@ def test_parse_config_ignores_unknowns(tmp_path, opt_manager, caplog):
...
+    with tmpdir.as_cwd():
+        cfg, _ = config.load_config("setup.cfg", [], isolated=False)
+
+    with pytest.raises(ValueError) as excinfo:
+        config.parse_config(opt_manager, cfg, tmpdir)
+
+    expected = (
+        f"Error code //comment "
+        f"supplied to '{option}' option "
+        f"does not match {VALID_CODE_PREFIX.pattern}"
```
asottile, line 236: hmm maybe my feedback didn't land -- there shouldn't be interpolation in tests

tests/unit/test_options_config.py
```diff
@@ -218,3 +218,35 @@ def test_parse_config_ignores_unknowns(tmp_path, opt_manager, caplog):
...
+
+    with pytest.raises(ValueError) as excinfo:
+        config.parse_config(opt_manager, cfg, tmpdir)
+
+    expected = (
+        "Error code //comment supplied to 'ignore' option "
+        "does not match ^[A-Z]{1,3}[0-9]{0,3}$"
+    )
+    msg = excinfo.value.args[0]
+    assert msg == (expected)
```
asottile, line 236: why didn't you use the code I wrote for you in the comment?  this assertion is weaker than the one I wrote for you as it allows an exception with more arguments to pass silently

also these parens do nothing

asottile (commented):
can you show the commandline experience when this fails?

tests/unit/test_options_config.py
mennoliefstingh: Hi, sorry about being quiet for a while but I caught a nasty covid case. Fixed this assertion now!

mennoliefstingh:
> can you show the commandline experience when this fails?

Error being raised with a dodgy .flake8 file: 

`flake8
  File "/usr/local/bin/flake8", line 33, in <module>
    sys.exit(load_entry_point('flake8', 'console_scripts', 'flake8')())
  File "/Users/menno/dev/flake8/src/flake8/main/cli.py", line 23, in main
    app.run(argv)
  File "/Users/menno/dev/flake8/src/flake8/main/application.py", line 333, in run
    self._run(argv)
  File "/Users/menno/dev/flake8/src/flake8/main/application.py", line 321, in _run
    self.initialize(argv)
  File "/Users/menno/dev/flake8/src/flake8/main/application.py", line 306, in initialize
    self.parse_configuration_and_cli(cfg, cfg_dir, remaining_args)
  File "/Users/menno/dev/flake8/src/flake8/main/application.py", line 148, in parse_configuration_and_cli
    self.options = aggregator.aggregate_options(
  File "/Users/menno/dev/flake8/src/flake8/options/aggregator.py", line 30, in aggregate_options
    parsed_config = config.parse_config(manager, cfg, cfg_dir)
  File "/Users/menno/dev/flake8/src/flake8/options/config.py", line 128, in parse_config
    raise ValueError(
ValueError: Error code // supplied to 'ignore' option does not match ^[A-Z]{1,3}[0-9]{0,3}$`

asottile (approved):

PR #1716, opened 2022-10-17
Document the types of plugins and expected results

jdevera:
# 📖 What is covered in this PR

I have recently attempted to write a plugin for Flake8. I found the video tutorial to be a great source of info, but understand it is desirable to move more and more of its contents to written form.

This PR is my attempt at moving some of that information to the docs.

Note this is based on my understanding after watching the video tutorial and having read the code to understand the different types of plugins, how they are loaded, how they are called, what parameters they get and what Flake8 expects them to return.

# 🙊 One notable omission

When describing the physical line plugins, I have deliberately omitted the fact that it supports receiving a single tuple from the plugin. I am basing this omission on my intuition that this is a special case to deal with legacy, one that Flake8 might not want to promote.

### DIFF
--- a/docs/source/plugin-development/checkers.rst
+++ b/docs/source/plugin-development/checkers.rst
@@ -1,0 +1,106 @@
+.. _checker-plugins:
+
+===============
+Checker Plugins
+===============
+
+A |Flake8| checker plugin needs to have a registered entry point. This entry
+point must reference a callable object.
+
+
+Declaring the type of plugin
+============================
+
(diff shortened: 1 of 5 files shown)


asottile:
hi -- appreciate the work you put into this but it breaks the links and I don't think it really makes things "better" -- the emoji in the description also make this PR look like spam so maybe keep that in mind for the future

asottile:
any time content is moved or pages are split or headings are changed the links break -- also this info is already in docs/source/plugin-development/plugin-parameters.rst

PR #1723, opened 2022-10-27
enable multiprocessing on other platforms

asottile:
I haven't actually tried this on windows or macos yet, but in theory it works

it also makes flake8 5-7% faster on `fork` platforms as well

### DIFF
--- a/tests/unit/test_legacy_api.py
+++ b/tests/unit/test_legacy_api.py
@@ -1,57 +1,12 @@
 """Tests for Flake8's legacy API."""
 from __future__ import annotations
 
-import argparse
-import configparser
-import os.path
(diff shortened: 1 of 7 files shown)


tests/unit/test_legacy_api.py
```diff
@@ -1,57 +1,12 @@
...
 from unittest import mock
 
 import pytest
 
 from flake8.api import legacy as api
 from flake8.formatting import base as formatter
-from flake8.options import config
-
-
-def test_get_style_guide():
```
asottile, line 16: I deleted this unit test since it's just a copy paste of the implementation -- there's an integration test which covers the same pathway

Issue #1729, opened 2022-11-03
--append-config: should error if file does not exist


asottile:
### describe the request

similar to #1498 -- this should also error for `--append-config`

```console
$ flake8 --config does-not-exist setup.py; echo $?
There was a critical error during execution of Flake8:
The specified config file does not exist: does-not-exist
1
$ flake8 --append-config does-not-exist setup.py; echo $?
0
```

LeetaoGoooo:
https://github.com/PyCQA/flake8/blob/ff8988bd586a4af192390e4316772333f34ba1e5/src/flake8/options/config.py#L89

Just raise an ExecutionError?

asottile:
yep, this should be very easy

LeetaoGoooo:
> yep, this should be very easy

i saw there was a todo 

https://github.com/PyCQA/flake8/blob/ff8988bd586a4af192390e4316772333f34ba1e5/src/flake8/options/config.py#L86

should we fix it ,meanwhile refactor the code?

asottile:
that todo is not simple to fix

LeetaoGoooo:
> that todo is not simple to fix

ok 😂 let me fix the simple one, i'll make a pr later

PR #1732, opened 2022-11-06
Raise a exception if append-config files does not exist

LeetaoGoooo:
This PR resolves https://github.com/PyCQA/flake8/issues/1729 by raising  a exception if append-config files does not exist

### DIFF
--- a/tests/unit/test_options_config.py
+++ b/tests/unit/test_options_config.py
@@ -222,6 +222,11 @@ def test_load_config_missing_file_raises_exception(capsys):
         config.load_config("foo.cfg", [])
 
 
+def test_load_config_missing_append_config_raise_exception():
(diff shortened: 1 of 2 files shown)


tests/unit/test_options_config.py
```diff
@@ -141,6 +141,12 @@ def test_load_config_append_config(tmpdir):
     assert cfg_dir == str(tmpdir)
 
 
+def test_load_config_no_append_config_found_raise_execution_error():
+    extra = "dont_exist_config.cfg"
+    with pytest.raises(exceptions.ExecutionError):
+        config.load_config(None, [str(extra)], isolated=False)
```
asottile, line 147: this should be next to the other test which checks the same thing

asottile (approved):

PR #1740, opened 2022-11-18
allow importlib-metadata 5 if on py3.7

dill0wn:
I know there have been countless issues on this matter, but here is a proposed fix.

The current release version of flake8 (5.0.4) gets into a dependency fight if you try to pin importlib-metadata>=5 (or if you have another package that needs it).

To fix this, I have proposed this changeset to allow flake8 to use importlib-metadata 5 on py37 only.


asottile:
the current dependency spec is for python3.7 only already so this just simply breaks it

PR #1772, opened 2022-12-20
allow idiomatic setup.cfg comments

bukzor-sentryio:


asottile:
in line with the [recommendations made in `configparser`](https://docs.python.org/3/library/configparser.html#customizing-parser-behaviour) (especially since it is a filed shared with other parsers) we intentionally do not support comments -- this is [documented here](https://flake8.pycqa.org/en/latest/user/configuration.html) (^F inline comments) and there are many duplicates in the issue tracker

bukzor-sentryio:
That would make for a much improved error message. The current error doesn't quite convey the same information:

```
$ flake8
  File "flake8", line 8, in <module>
    sys.exit(main())
  File "flake8/main/cli.py", line 23, in main
    app.run(argv)
  File "flake8/main/application.py", line 198, in run
    self._run(argv)
  File "flake8/main/application.py", line 186, in _run
    self.initialize(argv)
  File "flake8/main/application.py", line 165, in initialize
    self.plugins, self.options = parse_args(argv)
  File "flake8/options/parse_args.py", line 53, in parse_args
    opts = aggregator.aggregate_options(option_manager, cfg, cfg_dir, rest)
  File "flake8/options/aggregator.py", line 30, in aggregate_options
    parsed_config = config.parse_config(manager, cfg, cfg_dir)
  File "flake8/options/config.py", line 131, in parse_config
    raise ValueError(
ValueError: Error code '#' supplied to 'extend-ignore' option does not match '^[A-Z]{1,3}[0-9]{0,3}$'
```

The changelog does mention "incorrect error codes are rejected" but doesn't mention that comments are now intentionally unsupported, nor what to do about it. As such I inferred (incorrectly) that this was an unintentional side-effect.

asottile:
they were always unsupported but they (and each word in their condents) were *silently* being added to the ignore list

PR #1773, opened 2022-12-21
Enable linting of stub files (`.pyi`) in pre-commit hook

Apakottur:
Enable support for `pyi` files when using `flake8` through `pre-commit`. This will run flake8 on stub files in addition to source files. 

Note that by default flake8 doesn't run checks on `pyi` files, and most commonly the [flake8-pyi plugin](https://pypi.org/project/flake8-pyi/) is used to enable those files types and check them. However, without the change in the pre-commit config file, the stub files are still ignored.

References:
1. [The same file on black](https://github.com/psf/black/blob/main/.pre-commit-hooks.yaml)
2. [pre-commit docs](https://github.com/psf/black/blob/main/.pre-commit-hooks.yaml)
3. [Same PR on pyupgrade](https://github.com/asottile/pyupgrade/pull/771)
4. [Same PR on absolufy-imports](https://github.com/MarcoGorelli/absolufy-imports/pull/77)

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,5 +3,6 @@
     description: '`flake8` is a command-line utility for enforcing style consistency across Python projects.'
     entry: flake8
     language: python
-    types: [python]
+    types_or: [python, pyi]
     require_serial: true
+    minimum_pre_commit_version: 2.9.2


sigmavirus24 (commented):
This is wrong. Given flake8 doesn't support these natively and won't, this will cause more trouble than not for people using this hook as I understand it. But @asottile wrote pre-commit and pyupgrade and maintains this project so I'm sure he'll know the right call here.

FYI bombing a bunch of projects with identical PRs and nearly identical descriptions looks like spam. Linking to other unmerged PRs looks like you're going someone won't notice they're unmerged and think that they hold credibility and seems disingenuous. Finally, if the person who created and maintains there if the projects you're interacting with, maybe assume they've configured things as they should be

Apakottur:
@sigmavirus24 Thanks for the thorough reply.

Good point regarding this looking as spam, my thinking was that since this is the same proposed change for multiple projects, any comments on one PR might be relevant for someone looking at another, but in hindsight I agree with you about that.

I tested out flake8 with this change and it just ignores pyi files without the pyi plugin, but sounds like I'm missing something and maybe my PR is wrong, or simply undesired.

How would you suggest I approach a situation like this in the future, open an issue with a question/suggestion instead of a PR?

Please feel free to close this PR if it's undesired.

asottile:
flake8 itself doesn't support linting of non python files. flake8-pyi is technically outside of support

but the nice thing about pre-commit is you don't need this in flake8's config. any project can override the settings as needed:

```yaml
    -   id: flake8
        types: [file]
        types_or: [python, pyi]
        additional_dependencies: [flake8-pyi]
```

so rather than make this worse for everyone for an unsupported use case, one should do that instead

Issue #1777, opened 2022-12-30
Change valid code regexp


Dimitryliss1:
### describe the request

Current version's regular expression (defined [here](https://github.com/PyCQA/flake8/blob/main/src/flake8/defaults.py#L46)) doesn't allow us to use any linter, that uses 4-or-more digit error codes, as a plugin for Flake8 (For example, PyLint uses 4-digit codes, e.g. [here](https://pylint.pycqa.org/en/latest/user_guide/messages/messages_overview.html)).

So now this regular expression blocks us from updating to newer versions of Flake8.

Idea is to change this RE to, for example:
```
^[A-Z]{1,3}[0-9]{0,4}$
```

or even
```
^[A-Z]{1,3}[0-9]*$
```
allowing for more custom linters to be used.

asottile:
please search the issue tracker, there are many duplicates and this is not changing

PR #1782, opened 2023-01-10
Update VALID_CODE_PREFIX

Dimitryliss1:
Change VALID_CODE_PREFIX variable to allow codes with 4 digits.

With this change we can use plugins (such as [this](https://github.com/orsinium-labs/flake8-pylint) one) to connect Flake8 and PyLint, which uses message codes described above (examples here https://pylint.pycqa.org/en/latest/user_guide/messages/messages_overview.html)

Refers to (and fixes) #1777


asottile:
what part of "this is not changing" do you not understand?

PR #1783, opened 2023-01-10
add allow_no_value=True for RawConfigParser

phubaba:
Some plugins use arguments without values such as https://pypi.org/project/flake8-unused-arguments/

to be able to put the arguments into the config file we need to set allow_no_value=True


asottile:
we can't do this, it would break reading configuration for other tools

please go into more detail in an issue **first** about the concrete problem and use case

asottile:
all I see is `store_true` which is represented as a boolean in the configuration

PR #1799, opened 2023-02-17
Two links to important sites were added to the README

estuardodev:
Two links were added to the README.rst document that directs to the official PyPI site and frequently asked questions about pycqa for added security for users using this project.

This is to prevent new users to the project from ending up on fake sites that can index search engines.


asottile:
this looks like spam

no thanks

PR #1812, opened 2023-03-18
update ranges for dependencies

sadikkuzu:
https://github.com/PyCQA/flake8/blob/45699b61ae4522864c60480c80d7c2ffb952d52e/setup.cfg#L36-L39


sadikkuzu:
Sorry, I've seen `roughly as` expression later on.

PR #1818, opened 2023-04-07
feat: implement github format

goatwu1993:
implement [GitHub annotation](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message)

Motivation:

user can run flake8 --format=github and see github pr error annotation.


asottile:
you can make a plugin for this instead

PR #1827, opened 2023-05-09
Update configuration.rst

dmitry-kabanov:
Clarify information about how the search for configuration files is done, particularly, state that only the first found file is read for configuration.


asottile:
this isn't correct and is spelled out below how it actually works (the file with the flake8 section stuff)

PR #1839, opened 2023-06-28
Exclude virtual env dirs by default

paduszyk:
Hi! ✋🏻 

There are some linters/formatters which automatically exclude/ignore names commonly used for virtual enviroment dirs, namely: `env`, `.env`, `venv`, `.venv`. I guess, it would be nice to have this feature in Flake8 as well.

Thanks for considering! 🤝

### DIFF
--- a/docs/source/user/options.rst
+++ b/docs/source/user/options.rst
@@ -234,7 +234,7 @@ Options and their Descriptions
 
     Provide a comma-separated list of glob patterns to exclude from checks.
 
-    This defaults to: ``.svn,CVS,.bzr,.hg,.git,__pycache__,.tox,.nox,.eggs,*.egg``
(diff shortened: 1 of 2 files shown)


asottile:
there's no standard name and I don't want to entertain PRs from everyone's personal spelling of the directory

deronnax:
@asottile PDM and Poetry both creates the virtualenv in the project root under the `.venv` name. (relevant documentation page for [Poetry](https://python-poetry.org/docs/configuration/#virtualenvsin-project) and [PDM](https://pdm-project.org/en/latest/usage/venv/#virtualenv-auto-creation)). Can we at least the `.venv` excluded then?
[Ruff also excludes it](https://github.com/astral-sh/ruff?tab=readme-ov-file#configuration).

asottile:
please see the comment directly above yours.

PR #1852, opened 2023-07-29
use setup-cfg-fmt

asottile:


setup.cfg
```diff
@@ -27,18 +27,13 @@ classifiers =
 
 [options]
 packages = find:
-package_dir =
-    =src
-# We document the reasoning for using ranges here:
-# https://flake8.pycqa.org/en/latest/faq.html#why-does-flake8-use-ranges-for-its-dependencies
-# And in which releases we will update those ranges here:
-# https://flake8.pycqa.org/en/latest/internal/releases.html#releasing-flake8
```
sigmavirus24, line 35: Probably want to keep the comments here, no?

asottile: setup-cfg-fmt doesn't support comments so I'm fine killing them to get the convenience.  the blame should have sufficient information 🤷

PR #1867, opened 2023-09-09
Document the no support for pyproject.toml

AndreMiras:
Maybe link to the plugin supporting it within the documentation? https://github.com/john-hen/Flake8-pyproject

This PR is because it took me a bit of time to figure out flake8 didn't have support for pyproject.toml as there's no reference to it in the documentation


asottile:
there's no reason to document every unsupported thing when there's an exhaustive list of the supported things

PR #1898, opened 2023-12-05
Fix linting and test in CI

kurtmckee:
I ran `tox -e linters` locally on `main` and discovered that flake8-bugbear was reporting B018 ("useless Tuple expression") errors for two files.

This PR introduces the following changes:

* Resolve the flake8-bugbear B018 errors
* Add a config to run the linters in CI

This should help reduce the possibility that lint errors slip into `main`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,6 +7,16 @@ on:
   pull_request:
 
 jobs:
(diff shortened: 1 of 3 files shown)


asottile:
our linting is already done in CI if you scroll down

PR #1899, opened 2023-12-05
Fix a workflow schema violation

kurtmckee:
The `main.yml` workflow contains a schema violation; its `tags` property is `null` (must an array with at least one item).

This PR introduces the following changes:

* Fix a schema violation in `main.yml`
* Add a pre-commit hook to validate workflow and the ReadTheDocs file for schema violations

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,7 +3,6 @@ name: main
 on:
   push:
     branches: [main]
-    tags:
   pull_request:
 
 jobs:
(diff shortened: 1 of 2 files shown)


asottile:
whatever schema you're using is incorrect.  `tags: null` is "all tags"

kurtmckee:
That doesn't appear to be [documented](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore).

Would you open to me updating an alternate PR to use this syntax?

```yaml
tags:
  - "*"
```

asottile:
that would be a waste of both of our time

PR #1907, opened 2024-01-05
update plugins for sphinx-prompt-1.8.0

mgorny:
The sphinx-prompt plugin has renamed its package in 1.8.0 from erraneous `sphinx-prompt` name to `sphinx_prompt`.  Adjust the conf accordingly.


mgorny:
Thanks!

PR #1911, opened 2024-01-09
Add `project_urls` metadata

Viicos:
Let me know if you'd like to see more of them added


intgr:
FWIW from asottile's other repo: https://github.com/PyCQA/pyflakes/pull/781#issuecomment-1657965207

Viicos:
Same as you, didn't think this would have been declined. Thanks for letting me know

PR #1918, opened 2024-02-13
Add codespell configuration, pre-commit hook + make it fix some typos

yarikoptic:
TODOs
- [x] check if custom action needed or pre-commit will pick up new typo (removed now - pre-commit is enough)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,3 +39,8 @@ repos:
     hooks:
     -   id: mypy
         exclude: ^(docs/|example-plugin/)
+-   repo: https://github.com/codespell-project/codespell
+    # Configuration for codespell is in setup.cfg
+    rev: v2.2.6
+    hooks:
+    -   id: codespell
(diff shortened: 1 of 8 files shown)


.pre-commit-config.yaml
```diff
@@ -39,3 +39,8 @@ repos:
     hooks:
     -   id: mypy
         exclude: ^(docs/|example-plugin/)
+-   repo: https://github.com/codespell-project/codespell
+    # Configuration for codespell is in setup.cfg
+    rev: v2.2.6
+    hooks:
+    -   id: codespell
```
yarikoptic, line 46: note -- this is just a check, not a fixer, so if false positive - just add a skip in setup.cfg; i.e. there would be no bugs introduced by codespell

asottile:
no thanks, please open an issue before changing someone else's infra in the future

yarikoptic:
brr... ok. cherish the typos.

PR #1922, opened 2024-03-07
Make the source of E###, W### and C901 more explicit

peterjc:
It was all to easy to overlook the link to pycodestyle when it was under the table of F### codes.

### DIFF
--- a/docs/source/user/error-codes.rst
+++ b/docs/source/user/error-codes.rst
@@ -8,8 +8,15 @@ Flake8 and its plugins assign a code to each message that we refer to as an
 :term:`error code` (or :term:`violation`). Most plugins will list their error
 codes in their documentation or README.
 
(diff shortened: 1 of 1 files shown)


asottile:
the issue form and template are both explicit about this already

peterjc:
https://github.com/PyCQA/flake8/blob/main/.github/ISSUE_TEMPLATE/config.yml is clear - perhaps reorder it as I missed the ``E__`` code bit. Twice. I was probably rushed, sorry.

The pull request was to improve https://flake8.pycqa.org/en/latest/user/error-codes.html#error-violation-codes which doesn't mention ``E`` or ``W`` codes explicitly:

asottile:
unfortunately we have no control over the order there.  github used to put our links at the top and not any more

PR #1923, opened 2024-03-14
Update documentation regarding entry points

### LINKED ISSUE #1920: Update plugin registration documentation about entry points
### describe the request

Hello,

The documentation currently [states](https://flake8.pycqa.org/en/latest/plugin-development/registering-plugins.html#registering-a-plugin-with-flake8):

> **Flake8** relies on functionality provided by setuptools called [Entry Points](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#entry-points). These allow any package to register a plugin with **Flake8** via that package’s `setup.py` file.

Any interest in having this part of the documentation updated to point to the [Packaging user guide](https://packaging.python.org/en/latest/specifications/entry-points/) instead, so that people can be aware this is not a setuptools-only feature anymore?

(Without changing the setup.py example, I believe it is still up to date?)

Viicos:
Fixes #1920.

### DIFF
--- a/docs/source/plugin-development/index.rst
+++ b/docs/source/plugin-development/index.rst
@@ -30,7 +30,8 @@ To get started writing a |Flake8| :term:`plugin` you first need:
 
 Once you've gathered these things, you can get started.
 
-All plugins for |Flake8| must be registered via `entry points`_. In this
+All plugins for |Flake8| must be registered via
+:external:doc:`entry points<specifications/entry-points>`. In this
 section we cover:
 
 - How to register your plugin so |Flake8| can find it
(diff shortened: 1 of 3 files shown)


docs/source/plugin-development/index.rst
```diff
@@ -30,7 +30,8 @@ To get started writing a |Flake8| :term:`plugin` you first need:
 
 Once you've gathered these things, you can get started.
 
-All plugins for |Flake8| must be registered via `entry points`_. In this
+All plugins for |Flake8| must be registered via
+:external:doc:`entry points<specifications/entry-points>`. In this
```
sigmavirus24, line 34: Please user the `:external+<name>:doc:...` syntax for this, e.g, `:external+setuptools:`

sigmavirus24 (changes requested):
I think it will be better to be explicit about where we expect these docs to come from. If the links break, then it will be very confusing to debug *which* intersphinx mapping it is meant to come from and which one to look for to fix it

sigmavirus24:
Thank you @Viicos

Viicos:
Thanks for wrapping up

PR #1946, opened 2024-07-23
Fix wording of plugins documentation

Viicos:
Seems like I messed up in https://github.com/PyCQA/flake8/pull/1923


sigmavirus24 (approved):

sigmavirus24:
Thanks for the fix!

PR #1949, opened 2024-08-01
Handle escaped braces in f-strings

### LINKED ISSUE #1948: `logical_line` for call with f-string can result in invalid ast under Python 3.12

```console
$ pip install flake8
```

```json
{}
```

#### what I expected to happen

Some of the plugins in [`hacking`](https://pypi.org/project/hacking) use a combo of `ast.parse` and `logical_line` to generate an AST for an individual line that is then fed into an `ast.NodeVisitor` subclass. These work just fine on Python 3.11 and earlier, but I've noticed they fail under specific circumstances on Python 3.12. I've included a minimal reproducer below. There's a chance I am "holding it wrong" but I'd like to confirm this first.

#### sample code

`test.py`

```python
import unittest


class TestFoo(unittest.TestCase):

    def test_foo(self):
        response_key = 'foo'
        response_val = 'bar'
        body_output = ''
        self.assertEqual(
            f'{{"{response_key}": "{response_val}"}}', body_output
        )
```

`checks.py`:

```
class NoneArgChecker(ast.NodeVisitor):
    def __init__(self, func_name, num_args=2):
        self.func_name = func_name
        self.num_args = num_args
        self.none_found = False

    def visit_Call(self, node):
        # snipped
        ...


def foo(logical_line, noqa):
    if noqa:
        return

    for func_name in (
        'assertEqual', 'assertIs', 'assertNotEqual', 'assertIsNot'
    ):
        try:
            start = logical_line.index('.%s(' % func_name) + 1
        except ValueError:
            continue
        checker = NoneArgChecker(func_name)
        # print(logical_line)
        checker.visit(ast.parse(logical_line))
        continue
        if checker.none_found:
            yield start, "H203: Use assertIs(Not)None to check for None"
```

`tox.ini`:

```ini
[flake8:local-plugins]
extension =
    X123 = checks:foo
paths = .
```

#### commands ran

```console
$ flake8 test.py
wow.py:1:23: E999 SyntaxError: invalid syntax. Perhaps you forgot a comma?
```

If I comment out the `print` statement, I see different results for Python 3.12 compared Pythons 3.10 and 3.11. Under 3.12:

```
self.assertEqual(f'x{x{response_key}xxxx{response_val}xx}', body_output)
```

Under 3.10 and 3.11:

```
self.assertEqual(f'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx', body_output)
```

### additional information

The E999 error is associated with line 1 of the file, despite the error actually coming from a plugin. This led me on a wild goose chase as I tried to figure out why flake8 thought my file was invalid Python but Python itself did not. I suspect this might also be user error and I should be handling the potential exception from `ast.parse` in my plugin but again, I just wanted to confirm that this was expected practice.

stephenfin:
To use a curly brace in an f-string, you must escape it. For example:

```python
>>> k = 1
>>> f'{{{k}'
'{1'
```

Saving this as a script and running the 'tokenize' module highlights something odd around the counting of tokens:

```console
$ python -m tokenize test.py
0,0-0,0:            ENCODING       'utf-8'
1,0-1,1:            NAME           'k'
1,2-1,3:            OP             '='
1,4-1,5:            NUMBER         '1'
1,5-1,6:            NEWLINE        '\n'
2,0-2,2:            FSTRING_START  "f'"
2,2-2,3:            FSTRING_MIDDLE '{'     # <-- here...
2,4-2,5:            OP             '{'     # <-- and here
2,5-2,6:            NAME           'k'
2,6-2,7:            OP             '}'
2,7-2,8:            FSTRING_END    "'"
2,8-2,9:            NEWLINE        '\n'
3,0-3,0:            ENDMARKER      ''
```

The `FSTRING_MIDDLE` character we have is the escaped/post-parse single curly brace rather than the raw double curly brace, however, while our end index of this token accounts for the parsed form, the start index of the next token does not (put another way, it jumps from 3 -> 4). This triggers some existing, unrelated code that we need to bypass. Do just that.

Closes: #1948

### DIFF
--- a/src/flake8/processor.py
+++ b/src/flake8/processor.py
@@ -203,7 +203,13 @@ class FileProcessor:
             if token_type == tokenize.STRING:
                 text = mutate_string(text)
             elif token_type == FSTRING_MIDDLE:  # pragma: >=3.12 cover
-                text = "x" * len(text)
+                # A curly brace in an FSTRING_MIDDLE token must be an escaped
+                # curly brace. Both 'text' and 'end' will account for the
+                # escaped version of the token (i.e. a single brace) rather
+                # than the raw double brace version, so we must counteract this
+                brace_offset = text.count("{") + text.count("}")
+                text = "x" * (len(text) + brace_offset)
+                end = (end[0], end[1] + brace_offset)
             if previous_row:
                 (start_row, start_column) = start
                 if previous_row != start_row:
--- a/tests/integration/test_plugins.py
+++ b/tests/integration/test_plugins.py
@@ -1,6 +1,8 @@
 """Integration tests for plugin loading."""
 from __future__ import annotations
 
+import sys
+
 import pytest
 
 from flake8.main.cli import main
@@ -258,6 +260,39 @@ f'hello world'
 
     expected = """\
 t.py:1:1: T001 "f'xxxxxxxxxxx'"
+"""
+    out, err = capsys.readouterr()
+    assert out == expected
+
+
+def test_escaping_of_fstrings_in_string_redacter(tmpdir, capsys):
+    cfg_s = f"""\
+[flake8]
+extend-ignore = F
+[flake8:local-plugins]
+extension =
+    T = {yields_logical_line.__module__}:{yields_logical_line.__name__}
+"""
+
+    cfg = tmpdir.join("tox.ini")
+    cfg.write(cfg_s)
+
+    src = """\
+f'{{"{hello}": "{world}"}}'
+"""
+    t_py = tmpdir.join("t.py")
+    t_py.write_binary(src.encode())
+
+    with tmpdir.as_cwd():
+        assert main(("t.py", "--config", str(cfg))) == 1
+
+    if sys.version_info >= (3, 12):  # pragma: >=3.12 cover
+        expected = """\
+t.py:1:1: T001 "f'xxx{hello}xxxx{world}xxx'"
+"""
+    else:  # pragma: <3.12 cover
+        expected = """\
+t.py:1:1: T001 "f'xxxxxxxxxxxxxxxxxxxxxxxx'"
 """
     out, err = capsys.readouterr()
     assert out == expected


stephenfin:
As noted in #1948, this is mainly preventing this code for executing:

https://github.com/PyCQA/flake8/blob/2a811cc4d2aaed3e8eb5a9f04f08ccc8af7c0791/src/flake8/processor.py#L207-L218

However, I have no idea what that code is intended for nor if I can remove it outright. Hopefully someone here can enlighten me so I can stick a comment in for future contributors/future me.

src/flake8/processor.py
```diff
@@ -203,9 +203,17 @@ def build_logical_line_tokens(self) -> _Logical:  # noqa: C901
...
             elif token_type == FSTRING_MIDDLE:  # pragma: >=3.12 cover
-                text = "x" * len(text)
+                # A curly brace in an FSTRING_MIDDLE token must be an escaped
+                # curly brace. Both 'text' and 'end' will account for the
+                # escaped version of the token (i.e. a single brace) rather
+                # than the raw double brace version, so we must counteract this
+                fstring_offset = 0
+                if "{" in text or "}" in text:
+                    fstring_offset = text.count("{") + text.count("}")
+                text = "x" * (len(text) + fstring_offset)
```
asottile, line 213: probably simpler:
```python
brace_offset = text.count('{') + text.count('}')
text = "x" * (len(text) + brace_offset)
end = (end[0], end[1] + brace_offset)
```

tests/integration/test_plugins.py
```diff
@@ -248,16 +250,20 @@ def test_logical_line_plugin(tmpdir, capsys):
     cfg.write(cfg_s)
 
     src = """\
-f'hello world'
+f'{{"{hello}": "{world}"}}'
```
asottile, line 253: please write a separate test demonstrating your problem instead of changing the meaning of this test

asottile (commented):
please rebase out unrelated formatting changes

tests/integration/test_plugins.py
asottile: I meant an entirely separate test -- with a name describing what's being tested (specifically the handling of curly braces in fstrings) rather than overloading the existing test

stephenfin: Seems a bit overkill, no? This accomplishes the same thing without significant additional runtime cost, and it's arguably a better test owing or the fact it mixes strings and "code". From your comment on the issue:

> nope! the method is meant to redact string contents and not code

I can straight up copy-paste 99% of the test, but it just seems a bit daft.

asottile: I've asked for it twice now

PR #1958, opened 2024-11-18
Easier plugin debugging by noting source filename of internal exceptions

Zac-HD:
I patched this in while trying to track down where in a rather large codebase https://github.com/python-trio/flake8-async/issues/322 was triggered, and then thought that it was useful enough and had so little downside that I should open this PR 🙂


asottile:
for debugging I'd recommend setting a breakpoint instead!

asottile:
> What's the downside of this?

- version specific code
- untested
- maybe (?) breaks the stacktrace due to the `AttributeError` (idk I'd have to check)
- the exception output is potentially confusion with the existing messages

Issue #1965, opened 2025-02-13
Flake8 hangs on a windows machine with > 61 procs when running on > 1 file


pylbert:
### how did you install flake8?

```console
D:\temp>uv venv flake8
Using CPython 3.10.15
Creating virtual environment at: flake8
Activate with: flake8\Scripts\activate

D:\temp>flake8\Scripts\activate

(flake8) D:\temp>uv pip install flake8
Using Python 3.10.15 environment at: flake8
Resolved 4 packages in 85ms
Prepared 4 packages in 110ms
░░░░░░░░░░░░░░░░░░░░ [0/4] Installing wheels...
Installed 4 packages in 115ms
 + flake8==7.1.1
 + mccabe==0.7.0
 + pycodestyle==2.12.1
 + pyflakes==3.2.0
```

### unmodified output of `flake8 --bug-report`

```json
{
  "platform": {
    "python_implementation": "CPython",
    "python_version": "3.10.15",
    "system": "Windows"
  },
  "plugins": [
    {
      "plugin": "mccabe",
      "version": "0.7.0"
    },
    {
      "plugin": "pycodestyle",
      "version": "2.12.1"
    },
    {
      "plugin": "pyflakes",
      "version": "3.2.0"
    }
  ],
  "version": "7.1.1"
}
```

#### what I expected to happen

Flake8 should run and exit with defaults from the command line

#### Runs/does not exit (hangs after printing stack trace):

```console
(flake8) D:\temp>flake8 testit0.py testit1.py
Exception in thread Thread-1 (_handle_workers):
Traceback (most recent call last):
  File "lib\threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "lib\threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "lib\multiprocessing\pool.py", line 522, in _handle_workers
    cls._wait_for_updates(current_sentinels, change_notifier)
  File "lib\multiprocessing\pool.py", line 502, in _wait_for_updates
    wait(sentinels, timeout=timeout)
  File "lib\multiprocessing\connection.py", line 879, in wait
    ready_handles = _exhaustive_wait(waithandle_to_obj.keys(), timeout)
  File "lib\multiprocessing\connection.py", line 811, in _exhaustive_wait
    res = _winapi.WaitForMultipleObjects(L, False, timeout)
ValueError: need at most 63 handles, got a sequence of length 194   # <-- hangs here
```

[...]

PR #1966, opened 2025-02-16
avoid starting unnecessary processes when file count is limited

asottile:


asottile:
this will make #1965 significantly less likely but does not actually "solve" that problem -- arguably #1965 if it is actually a problem should get solved in cpython and not here

Issue #324, opened 2021-04-03
Ignore warnings on block of lines


asottile:
In GitLab by @volopas on Nov 23, 2019, 08:37

When we add special comments with `noqa` keyword, we can ignore either an exact line of code, or the entire file, but there is no way to ignore several lines of code without being too verbose. Nor can one ignore some but not the other warnings on the entire file.

I like how eslint [tackles the whole issue](https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments). They have comments 1) able to ignore all at the same time or specific warnings; 2) the entire file, a code fragment, a current line or the one next to a current line. So their system is both flexible and complete.

The specific problem I have making me start the issue is that I have to have a bunch of lines with module-level `import`, not at the top of the file (E402). It looks like that:

```python
import sys
# and other top-level imports (1)

# some configuration needed
# to import other modules (2)

# the bunch of other imports
# in the configured environment (3)
```
I wouldn't like disabling E402 entirely because that's just a necessary exclusion. But I can only disable E402 warning either in every single line of the block (3) or in the entire file. Both choices seem ugly or missing the point.

It would be great if flake8 would get more flexible in this way. Thank you for your efforts and for an undoubtedly useful tool.

asottile:
In GitLab by @asottile on Nov 23, 2019, 09:59

I'd rather not if possible

Have you tried `extend-ignore` / `per-file-ignores`? It sounds like those would solve your problem just as well

asottile:
In GitLab by @volopas on Nov 23, 2019, 12:36

Thanks, so we can ignore some particular warnings on a per-file basis and use inheritance to extend the list down the hierarchy. That'll be definitely helpful for me in the future. It's a pity though we cannot mark a code fragment larger than a couple of lines and smaller than a file in a concise manner. I wouldn't like to ignore E402 entirely for the file, I'd rather follow it with this particular case excluded.

It seems you see some flaws in inline ignores that make them somewhat unwanted. What's your opinion on the matter?

asottile:
In GitLab by @asottile on Nov 25, 2019, 08:13

I'm mostly hesitant because there's already working solutions to this problem and implementing the suggested feature would be potentially complicated (and something we'd have to support for forever).  The inclusion/exclusion/ignore functionality is already some of the most complicated parts of `flake8`.

[...]

PR #1972, opened 2025-03-25
Add ability to disable qa for blocks of code

tomdottom:
Adds an ignore block which can be defined with `# noqa: on` and `# noqa: off` comments.

Ignore block behaviour should be the same as inline ignore.

Example:

```python
# noqa: on E501
really_long_one = "blah blah blah blah blah blah blah blah blah blah blah"
really_long_two = "blah blah blah blah blah blah blah blah blah blah blah"
really_long_foo = "blah blah blah blah blah blah blah blah blah blah blah"
# noqa: off
```

### DIFF
--- a/src/flake8/defaults.py
+++ b/src/flake8/defaults.py
@@ -40,6 +40,25 @@ NOQA_INLINE_REGEXP = re.compile(
     re.IGNORECASE,
 )
 
+NOQA_BLOCK_REGEXP = re.compile(
+    # We're looking for items that look like this:
+    # ``# noqa: off``
+    # ``# noqa:off``
+    # ``# noqa: Off``
+    # ``# noqa:Off``
+    # ``# noqa: off E123``
+    # ``# noqa: off E123,W451,F921``
+    # ``# noqa:off E123,W451,F921``
+    # ``# NoQA: off E123,W451,F921``
+    # ``# NOQA: off E123,W451,F921``
+    # ``# NOQA:off E123,W451,F921``
+    # We do not care about the casing of ``noqa``
+    # We want the desired block state
+    # We want a comma-separated list of errors
+    r"# noqa:[\s]?(?P<state>off|on)[\s]?(?P<codes>([A-Z]+[0-9]+(?:[,\s]+)?)+)?",
+    re.IGNORECASE,
+)
+
 NOQA_FILE = re.compile(r"\s*# flake8[:=]\s*noqa", re.I)
 
 VALID_CODE_PREFIX = re.compile("^[A-Z]{1,3}[0-9]{0,3}$", re.ASCII)
--- a/src/flake8/style_guide.py
+++ b/src/flake8/style_guide.py
@@ -423,8 +423,13 @@ class StyleGuide:
         error_is_selected = (
             self.should_report_error(error.code) is Decision.Selected
         )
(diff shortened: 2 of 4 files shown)


asottile:
there's a duplicate (rejected) issue

tomdottom:
Hi @asottile,

I assume you are referring to #324 from which I quote:

> if it becomes easier to implement in the future (I hope it will!) we'll consider this :)

Given this PR contains a **working and tested implementation** of the feature, and that it was quite simple to implement, I would be grateful if you could give some more details on the reasoning behind rejecting this PR.

I accept that there may be nuances and 2nd-order consequences which I don't fully appreciate.  But I don't see any described in the above referenced issue.

> I think inline ignores are fine as is for 99.9% of cases -- the case you're suggesting is particularly exceptional in my opinion (and as such I'm hesitant to implement a big complex feature for it)

I hope, once you've taken a quick look at the implementation, that you'll agree that it's not complex, and is basically just a specialisation of the `is_inline_ignored`.

If there's anything that's missing, needs to be cleaned up, or made clearer, just comment on the PR and I'll make the changes.

I would also disagree that this is an exceptional 0.01% need.  I regularly find myself wishing I could just disable a block instead of disabling a check for an entire file.  And it appears that the same feature request is made regularly by others.

asottile:
it makes some breaking changes to the comment format and significantly slows down flake8 so no it isn't just "simple"

tomdottom:
> ... it makes some breaking changes to the comment format

I did notice that I wasn't able support `on/off` block comments for formats like `# noqa, analysis:ignore`, but I didn't see any others.  What did I miss?

I've seen alternative suggestions like `# noqa-on / # noqa-off` which may cause less conflict or confusion if this would make the PR more acceptable to you.

> ... significantly slows down flake8

An understandable concern.  But you may be surprised how little overhead this adds.

Both `is_inline_ignored` and `is_block_ignored` rely on `linecache.getlines` (note that `getline` calls `getlines`) so there isn't any extra work done loading files into the linecache. 

Most of the extra work amounts to running a regex on each line.

I ran a quick test over the `cpython/` codebase to get some idea of the overhead:

Using `flake8 7.1.2`:

```shell
time flake8 --jobs=1 cpython/

....

real	2m39.925s
user	2m38.557s
sys	0m0.589s
```

Using this PR :

```shell
time flake8 --jobs=1 cpython/

....

real	2m40.507s
user	2m39.229s
sys	0m0.682s
```

Admittedly, the CPython codebase may not have may instances of `# noqa` , and certainly none of the new block format.  But the `is_block_ignored` will have been run on every file in `cpython/Lib/` (approx 2000 .py files), and a regex run on every line (approx 960k in .py files), with about 2s of overhead.

If I've missed anything, or got my wires crossed, please let me know.

T

asottile:
it's significantly worse on codebases with lots of errors, lots of noqa comments, or lots of blocks (I imagine you won't be able to find the latter one though)

there's also the question of "what to do about improper nesting?" "what about unclosed ons?" -- it doesn't seem like you've put much thought into either of these problems and as such made a pretty bad footgun

realistically I think we can't accept a feature like this and likely never will.  the comment you linked as motivation to work on this is a "maybe" and at least for now "maybe means no". 

I'm not here to debate this further -- please understand that it's likely flake8 will be in maintenance mode for quite some time until some major structural things are worked out.

PR #1990, opened 2025-07-07
Update hooks and use `autopep8` + `add-trailing-comma` instead of `black`

mxr:
`reorder-python-imports` and `black` are incompatible (https://github.com/psf/black/issues/4175). This blocks automatic `pre-commit` updates (https://github.com/PyCQA/flake8/pull/1912). As a workaround, replace `black` with `autopep8` and `add-trailing-comma`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,10 @@
 repos:
+-   repo: https://github.com/asottile/add-trailing-comma
+    rev: v3.2.0
+    hooks:
+    -   id: add-trailing-comma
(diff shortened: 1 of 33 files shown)


asottile:
I'm in favor, we'll see what @sigmavirus24 thinks  -- this at least would allow us to get updates.  we lose some of the automatic formatting that black decides though

maybe `add-trailing-comma` could get us the rest of the important parts though?

mxr:
@asottile @sigmavirus24 sounds like we are all on the same page. This PR is tidied up from my side now. Thank you!

asottile (approved):
(kitty cat script broken on this computer sadface)

PR #2023, opened 2026-05-01
Document how to test plugins

### LINKED ISSUE #705: Document how to test plugins
In GitLab by @stephenfin on May 23, 2019, 03:07

I'm trying to migrate [hacking](https://opendev.org/openstack/hacking) from [flake8 2.x to 3.x](https://review.opendev.org/#/c/651565/) and finding it extremely difficult to do so because the internals of the flake8 have changed so drastically between these versions (plus hacking does some stuff that's too smart for its own good, IMO). The flake8 documentation provides some introductory text on plugins, including a guide on [writing a "formatting type" plugin](http://flake8.pycqa.org/en/latest/plugin-development/formatters.html) (though it doesn't explain what a formatting plugin actually is...) but there's nothing about testing this or even what type of class or function the entrypoints should point to. The biggest issue I have is that I want to run flake8 in it entirety (i.e. with both the standard checks and my plugin enabled) on a test string, but there doesn't appear to be anyway to do this.

Would it be possible to get a small worked example that demonstrates:

- What the entrypoint should look like (i.e. if it can be a Python function, a class, or both)
- How you can test a string against only a single check
- How you can test a string against flake8's standard checks (without having to resort to a subprocess call)

I'll happily submit a documentation PR based on this example, if it would be helpful.

tanviranindo:
## Summary

Adds a **Testing plugins** page under the plugin development docs: editable installs, the legacy Python API (`get_style_guide` / `check_files`), temporary files (since `input_file` ignores `lines` on Flake8 3+), narrowing with `select`, and a subprocess CLI example.

## Related

Closes #705.

## Verification

`sphinx-build -E -W -c docs/source/ -b html docs/source/ docs/build/html` (docs tox env) succeeds locally.

### DIFF
--- a/docs/source/plugin-development/index.rst
+++ b/docs/source/plugin-development/index.rst
@@ -43,6 +43,8 @@ section we cover:
 
 - How to write your check plugin so that it works with |Flake8| 2.x and 3.x
 
+- How to test your plugin (:doc:`testing-plugins`)
+
 
 Video Tutorial
 ==============
@@ -65,3 +67,4 @@ Detailed Plugin Development Documentation
     registering-plugins
     plugin-parameters
     formatters
+    testing-plugins
--- a/docs/source/plugin-development/testing-plugins.rst
+++ b/docs/source/plugin-development/testing-plugins.rst
@@ -1,0 +1,91 @@
+=================
+ Testing plugins
+=================
+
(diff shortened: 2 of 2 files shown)


asottile:
this reads like ai slop.  I didn't think this improves things

